PR #1736, opened 2015-02-03
Changed default "skin-url" value

NotWoods:
With the new skin system introduced in 1.7.10/1.8, Mojang now hosts skins under the minecraft.net domain instead of Amazon's. Since the Amazon link now only provides outdated skins, it would be better to use the new skins.minecraft.net URL as default.


mikeprimm:
Looks good - thanks!

PR #2087, opened 2017-04-17
Move from mcstats.org to bstats.org for Metrics

KovuTheHusky:
It appears mcstats.org has been down for months and is probably not going to be maintained anymore. Most plugin authors seem to be moving to [bstats.org](https://bstats.org/) for Metrics.

Right now I have the dynmap plugin on bstats.org registered to myself but I would be happy to do a hand-over. I just didn't want some random person squatting on it.


KovuTheHusky:
And now it would appear that mcstats.org is back online after months of downtime. I'll leave this PR open for now just in case mcstats.org decides to go offline for an extended period of time again.

KovuTheHusky:
So mcstats.org is giving 500s and empty responses again - I'm personally sticking with bstats.org for my projects. I can update this PR to include both if you'd like but I think mcstats.org is on its way out in the minds of plugin developers.

mikeprimm:
Cool - I'll review and accept this ASAP - thanks!

PR #2188, opened 2018-01-17
Add support for LuckPerms offline permission checks

### LINKED ISSUE #2127: Players can use Webchat only when logged in to the minecraft server
Hello,
i have set `webchat-permissions: true`, and the permission check works, blocking chat lines when players do not inherit the dynmap.webchat, but for some strange reason players can enter lines that successfully publish on the bukkit chat only when they are actually connected to the minecraft server.

configuration.txt (here: https://hastebin.com/ilobabelix.vbs)

I wrote here because i cannot find anything to work around and make the chat work for users not logged in to the minecraft server.
The plugin version is: Dynmap version: core=2.5-Dev201706100401, plugin=2.5-Dev201706100405
The server version is: Paper version git-Paper-1152 (MC: 1.12) (Implementing API version 1.12-R0.1-SNAPSHOT)

thank you for help

lucko:
As requested here:
https://www.spigotmc.org/threads/luckperms-an-advanced-permissions-plugin.174259/page-91#post-2843682

Fixes #2127 

---

I've tested the change, all seems to be working correctly. Offline permission check runs ok.

https://gist.github.com/anonymous/fe56b86040c52a1f8b852db73926d7cb

The `Storage` interface in the LP API returns Futures - they can cause lag if joined from the server thread, however my understanding is that the offline check methods are only called from the web server anyway - so shouldn't ever be an issue.

I've tried to copy the existing code style - might not be 100% though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,6 +146,11 @@
 			<version>2.10.1</version>
 		</dependency>
 		<dependency>
+			<groupId>me.lucko.luckperms</groupId>
+			<artifactId>luckperms-api</artifactId>
+			<version>4.0</version>
+		</dependency>
+		<dependency>
 			<groupId>org.bstats</groupId>
 			<artifactId>bstats-bukkit</artifactId>
 			<version>1.1</version>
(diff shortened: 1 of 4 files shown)


mikeprimm:
Looks promising - thanks!

PR #2365, opened 2018-10-19
Fix loading profiles to solve 500 errors on login

haffmans:
This loads profiles from the server cache, which properly sets the uuid
(which again is used to check for bans in some other layers).

Backport of [DynmapForge PR 77](https://github.com/webbukkit/DynmapForge/pull/77)


mikeprimm:
Cool - thanks for the fix!

PR #2391, opened 2018-11-28
Added MySQL connection flags option "flags" in configuration.txt

aarroonn789:


aarroonn789:
@mikeprimm This adds a configuration option for MySQL connection flags. Newer versions of MySQL require you to explicitly state "useSSL=false" when connecting. If you do not, you get error messages in the Minecraft server console and the MySQL log each time DynMap interacts with the MySQL database. The hard coded flag "allowReconnect=true" prevents users from adding the "useSSL=false" flag to the database name. This change moves the hard coded flag to a default config option and allows the user to add the "useSSL=false" flag if their database requires it.

mikeprimm:
Looks good - thanks!

PR #2433, opened 2019-01-30
Add 1.13 biome mappings and colors

KovuTheHusky:
Add 1.13 biome mappings and colors (amidst scheme).


KovuTheHusky:
Will update for 1.14 as well once it's actually out. Bamboo biomes and such!

mikeprimm:
Looks good - thanks!

Issue #2415, opened 2019-01-11
Suggestion on how to reduce the transfered image size when using SQLite


alpe12:
## FEATURE REQUEST
- [ ] I would like to propose an additional feature.
- [ ] I would like to propose a config change.
- [ ] I would like to request mod support.
- [x] I would like to propose another change.

- [ ] I have searched the wiki and confirmed that this feature is not already implemented.

SQLite uses pages size of 4KB. That means that if a image stored has 5KB, it will actually use 8KB (2 pages of 4kb).
If the webpage has compression enabled, this is no problem, as compression takes care of that extra size. But if not (or using HTTPS) the full 8KB is sent unnecessarily.
What I have done (and recommend to be implemented) is to modify SQLite_tiles.php.

Replace this:
`echo $row[0];`

With this:
`echo rtrim($row[0], "\0");`

This trims the NUL characters at the end of image (they are not needed). Therefore, the exemple image now only transmit it's real size of 5KB instead of 8KB.
This is extremelly lightweight on CPU usage, hardly to make any changes even on very big servers, and saves KB on the transmitted image.
Since $row is already on memory, I don't think this will use more memory than it's already used.

(This may be useful on MySQL too).

Psy-Virus:
I think even if it would use a bit more RAM, it is a good trade-off.
3KB / picture saved times 500,000 pictures (whats quite common) is about 1.5 GB traffic that's saved.

For me in particular: I have about 797,684 png files with an average size of 11.86KB. So it would save me 109MB if I had a traffic plan and if all files were loaded once and if I would use an SQL based save method.

For JPEG files the amount of saved traffic would probably be higher.

Why don't you create a Pull request?

alpe12:
Good question. 
Did it now.

Psy-Virus:
:D great

alpe12:
https://github.com/webbukkit/dynmap/pull/2435

PR #2435, opened 2019-01-31
Reduce the transfered image size when using SQLite

alpe12:
https://github.com/webbukkit/dynmap/issues/2415


mikeprimm:
Looks good - I'm going to do a bigger change after this so that the DB has an ImageLen column to go with the Image - it'll be a new column, so it'll only have valid values as tiles are re-rendered, but it'll make it so that the length to trim to can just be returned vs being scanned for (even though the fix you did will work for PNG and JPG file both, since they have nonzero byte end markers).

Issue #2474, opened 2019-05-02
Required login bypass vulnerability


Ry0taK:
## Vulnerability Information
- Version: v3.0-beta-3 or below
- Type: [CWE-284](https://cwe.mitre.org/data/definitions/284.html) (Improper Access Control)

## Issue Description
Due to missing login check in `org.dynmap.servlet.MapStorageResourceHandler.java`, attacker can see map image without login despite "login-required" is enabled in configuration.

## Reproduce Steps
1. Enable `login-required` and `login-enabled` in configuration.
2. Try this [PoC](https://gist.github.com/Ry0taK/35342a13ef33b5afabee64718f61c85e)

Ry0taK:
[CVE-2019-12395](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-12395) has assigned to this vulnerability.

Ry0taK:
Hello @mikeprimm ,
I've sent to this vulnerability information to JVN (Japan Vulnerability Notes),
and JVN asked me when will the next release.
Can you tell me when will the next release please?
also, JVN said "CVE has already published, so if there is no plan for the next release, can you release next version as soon as possible please?".

Thanks,
RyotaK

mikeprimm:
It'll be released when 1.14.2 support is done - the issue here is so trivial (there are probably 3 or 4 servers of all the servers that use dynmap that even use the option of supporting login, restricting map visibility to using login AND using the internal web server) that I'm not driving a formal update just to release the fix.  I appreciate the fix, but I do feel that the opening of a CVE and JVN is making a 'mountain out of a molehill'.  It'll be released when the more important elements of the update are ready - hopefully this weekend.  The patched source is available and freely buildable, so anyone who needs it sooner is already has the fix available, as the mod is open source...

Ry0taK:
Thank you for your reply!

I forwarded your reply to JVN.
I understand that the severity of this vulnerability is low and I also understand that you don't release software just to fix this vulnerability.
However, CVE and JVN are vulnerability information databases, and the vulnerability severity does not matter.

Thanks,
RyotaK

mikeprimm:
I think we're looking good for being able to do a release on the latest code base this weekend - I finished up the remaining critical 1.14.x issues last night, and I THINK the latest dev build is a potential 'release candidate'.

Ry0taK:
Hello mikeprimm.

Thank you for the information.
I sent this information to JVN.
[...]

PR #2475, opened 2019-05-02
Fix required login bypass vulnerability

Ry0taK:
Patch for #2474


mikeprimm:
Looks good - thanks for fix!

PR #2498, opened 2019-06-01
Added functionality for restarting the scheduled job that stores markers

stormboomer:
This should fix problems with markers not being stored after a dynmap reload, as well as vanishing markers.
This only fixes the issue for the spigot version.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
@@ -140,7 +140,7 @@ public class DynmapCore implements DynmapCommonAPI {
     public boolean is_reload = false;
     public static boolean ignore_chunk_loads = false; /* Flag keep us from processing our own chunk loads */
 
-    public MarkerAPIImpl   markerapi;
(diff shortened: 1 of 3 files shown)


DynmapCore/src/main/java/org/dynmap/DynmapCore.java
```diff
@@ -140,7 +140,7 @@
     public boolean is_reload = false;
     public static boolean ignore_chunk_loads = false; /* Flag keep us from processing our own chunk loads */
 
-    private MarkerAPIImpl   markerapi;
+    public MarkerAPIImpl   markerapi;
```
VaiTon: I'm not sure about making this a public field. It could cause other plugin devs to access it instead of the main API.

stormboomer: you are probably right on that one. I will have a look into it and see if I can come up with a different solution, thats not changing too much of the core code.

mikeprimm:
Looks good - thanks for figuring this one out!

Issue #2533, opened 2019-06-29
1.14 markers vanishing [signs]


gravestristen:
Dynmap version: Dynmap-3.0-beta-4-spigot.jar
minecraft version: 1.14.2
server jar: spigot-1.14.2.jar
other plugins: worldedit-bukkit-7.0.0.jar
Config file: https://pastebin.com/bqebh6Cz
Problem: Dynmap markers made using signs disappear after rechecked by Dynmap,
description: I made a marker using a sign with the option enabled in the config, at first the marker stays on the map, but if i hang around the sign too long it'll disappear, if I immediately leave the area, IE go through a portal (I was using signs to mark the portals with Dynmap so they can easily be removed and updated) the marker will stay on the map, but if I re enter the area it'll disappear again, also after further investigating this also causes the marker to actually be removed from the marker.yml file. The format I have been using on the signs is.
`Line1: [dynmap]
Line2: name
Line3: set:portals
Line4: icon:portal`
The portals set has been created correctly.
Edit: formatted the sign example.

edelan7:
I also have the same issue on 1.14.3. Weirdly, one signs is permanent. If I add any others the marker.yml sometimes updates, sometimes doesn't ... as soon as I move away from the sign it disappears from the map and marker.yml. Other times I can add a sign, marker.yml gets updated, but once a third sign is added both the 2nd and 3rd sign disappear from the map. In all of the scenarios I've tested, the first sign never gets removed or deleted.

gravestristen:
I've also managed to make a permanent sign on accident and I had to remove it via commands, how i did it was i just wrote [dynmap] on it then exited the sign and deleted it because I didnt like the placement and I had to remove it via commands because it wouldnt go away.

Bernardez:
I also have this issue for the last few months while running different versions of Spigot 1.14(.3) and Dynmap snapshot releases and the latest beta Dynmap-3.0-beta-4-spigot.
I've only noticed markers 'vanishing' when players add new markers using a sign.

Luysita:
Same issue here.

steveabraham:
Just a +1 from me. I've been pulling my hair out trying to figure out what was wrong. I assumed it my set-up, since I use a standalone web server, but it seems not.

steveabraham:
Today's release doesn't seem to have sorted the issue for me. Some signs are now staying and persisting, some are vanishing after a few moments.

steveabraham:
[...]

PR #2584, opened 2019-08-14
Fix for disappearing sign-markers on 1.14 (Issue #2533)

tiagoad:
I've tested this on Paper 1.14, and included (untested) forge code that mimics the old code. I've never messed with Forge so I didn't even manage to get it to build, some looking into that will be needed, as well as testing with CB 1.11 (although I think it should probably work).


mikeprimm:
This looks promising - I need to review further and test on other platforms

mikeprimm:
Thanks for the PR!

Issue #2536, opened 2019-07-02
Missing API specification with 1.14.


mibby:
Running on 1.14, seems to need an API specification in the plugin.yml.

```
[13:22:52] [Server thread/WARN]: Plugin dynmap v3.0-SNAPSHOT-215 does not specify an api-version.
```

Meledosia:
I know this is a nuisance bug, but I also know it only takes 30 seconds to fix. Please add **version: 1.14** to the plugin.yml file so we don't have warnings on startup.

WeltensiedlerDE:
This is a bad idea. To keep plugins runnable on older versions, it's not recommended to add the AÜI string to the plugin.yml without creating several versions for different mc-versions.

mikeprimm:
We support old versions, and I'm not adding a new build just to make a message go away - sorry.  Feel free to fork and build your own version, though - just don't distribute it! :)

TheDGOfficial:
You can just add api-version: 1.13 to plugin.yml and it would not give warnings in 1.14, nor it will show plugin as dynmap* in the plugin list.

This does not break compatibility with pre-1.13 versions because there are no api-version setting in the old versions already (Bukkit ignores not existing plugin.yml settings like Java's MANIFEST.MF file). I just installed dynmap v3.0 beta5 and it showed a warning like this, and i just changed the api-version to 1.14 and it worked. (I didn't test the map, it just does not give startup errors and i assume it works)

I just want you to add this to plugin so no more users will get this warning. If you put api-version: 1.14 it will not work on 1.13, but if you add api-version: 1.13 it will work on every version. (unless you use newly added 1.13 APIs in older versions, or you are not using the 1.13 API because if api-version is not specified Bukkit applies special compatibility things that obviously may slow down load time)

The warning does not appear on 1.14 if you put api-version: 1.13 because nothing much is changed between (well many things changed in the background) 1.13 and 1.14. If spigot makes a new move and increases the api-version limit, then you will have to adapt it as well.

So, the downsides of not having the api-version is..

- It slows down the load time because Bukkit applies special mapping compatibilty thingy that reads, parses and processes class files

- In 1.14 (not 1.13 though) the plugin will give a warning like the issue specified in here - "Plugin x does not specify an api-version."

- The plugin will show extra * character at the end of the name in 1.13 or 1.14. This applies to every location, e.g: "Enabling dynmap*", "Disabling dynmap*", "Plugins(1): dynmap*" etc.

[...]

PR #2585, opened 2019-08-14
Fix missing api-version warning

TheDGOfficial:
As per comment from https://github.com/webbukkit/dynmap/issues/2536#issuecomment-521458380

### DIFF
--- a/spigot/src/main/resources/plugin.yml
+++ b/spigot/src/main/resources/plugin.yml
@@ -1,6 +1,7 @@
 name: dynmap
 main: org.dynmap.bukkit.DynmapPlugin
 version: "${version}-${buildnumber}"
+api-version: "1.13"
 authors: [mikeprimm]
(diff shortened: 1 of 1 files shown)


Nicofisi (approved):
Yes, we need that

mikeprimm:
Code technically supports versions older than 1.13 (and still has users using it) - need to confirm that this doesn't break those older versions.

TheDGOfficial:
> Code technically supports versions older than 1.13 (and still has users using it) - need to confirm that this doesn't break those older versions.

It does not break since 1.12 and older versions do not have an "api-version" thing and they don't recognize the "api-version" setting in the plugin.yml. Bukkit just ignores unrecognized settings in the plugin.yml.

mikeprimm:
Sounds good enough for me - thanks for the verification!

Issue #2124, opened 2017-07-16
HD skins support


prettydude:
My server uses patched Authlib, for the ability to use HD skins. But a transparent image is displayed on the map. 
I drew a square in place where the face should be, and map uses this square as face.
It would be great if the images were generated in proportion to the size of the skins.

Skin that i use for testing (yeah... i chose random skin from google)
![skin for testing](https://user-images.githubusercontent.com/19638254/28246579-010d1df6-6a26-11e7-8114-3d0c21c11b44.png)

Map rendered face
![viewformmap](https://user-images.githubusercontent.com/19638254/28246580-010db414-6a26-11e7-9fa8-1c38a3e3556e.png)

ExerciseBook:
Maybe I fixed this bug.
https://github.com/ExerciseBook/dynmap/commit/7f84262bfb1c05a782495d8a4427f4afa90b1095

I got an avatar form skin files in the original size.
After that, I scale this avatar in size of 8x8, 16x16 and 32x32.


After I tested it in the production environment. I will make a PR.

PR #2586, opened 2019-08-15
HD skins support

ExerciseBook:
https://github.com/webbukkit/dynmap/issues/2124

image


mikeprimm:
This looks promising - will review further.  Used to have problems with these functions on older headeless VMs (not functioning, memory leaks), which is why they were not used, but should be OK to use with our requirement for Java 8+

mikeprimm:
Thanks for the PR!

PR #2618, opened 2019-09-27
Add initial postgres database support.

baconwaifu:
Managed to modify the mySQL database backend for postgres, but since spigot doesn't pack the required JDBC driver like it does for mysql, either we need to download it at first run, bundle it in, or require the user go get it first. If it's decided to go with the option of "bundled" then someone else will have to tie that into gradle, since I don't know how. (personally, I'd have it download it at first configured run)

The internal webUI works perfectly, outside of markers. need to figure that out. Standalone isn't complete yet, but that should be some minor SQL syntax tweaks compared to the full blown type rewrite needed here.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
@@ -57,6 +57,7 @@ import org.dynmap.storage.filetree.FileTreeMapStorage;
 import org.dynmap.storage.mysql.MySQLMapStorage;
 import org.dynmap.storage.mariadb.MariaDBMapStorage;
 import org.dynmap.storage.sqllte.SQLiteMapStorage;
+import org.dynmap.storage.postgresql.PostgreSQLMapStorage;
 import org.dynmap.utils.BlockStep;
 import org.dynmap.utils.ImageIOManager;
 import org.dynmap.web.BanIPFilter;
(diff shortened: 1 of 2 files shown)


mikeprimm:
This looks really good - I need to do some additional review, and some testing

baconwaifu:
I've been running it on a live server since I made the fork, with the only problem I've encountered being that player positions would sometimes stop updating until you reload the page. that seems to be an unrelated bug, though.

mikeprimm:
Sounds good - any specific postgres version sensitivity you can think of?  It looks pretty generic, so I'm not expecting any problems (I work with PG on a routine basis at my job - just never got around to doing this port here :) )

baconwaifu:
Don't know of any DB version problems. Also, I just found the problem with markers. the marker json is being sent to the browser as a giant hex string; is also stored in the DB that way. Need to figure out if it's dynmap doing that, or postgres.

baconwaifu:
Looks to actually be jdbc. Dynmap calls `content.getBytes(UTF8)` and then stores the result, so JDBC is encoding it as hex to be safe. might change that column from TEXT to BYTEA, or remove the `GetBytes()` call and rely on TEXT being UTF-8 safe. (which it usually is, from what I can tell)
EDIT: dynmap treats it as a byte[] type for retrieval, so the column should be BYTEA.

mikeprimm:
Yep - might be better to go with BYTEA - it's just opaque data, as far as the DB is concerned, so no reason to worry about what code page the DB might default to

baconwaifu:
Change built, and it works. markers show up on the map again.

mikeprimm:
Merged - thanks!

PR #2623, opened 2019-10-03
fixed disappearing markers (webbukkit#2533), many thanks to tiagoad

### LINKED ISSUE #2533: 1.14 markers vanishing [signs]

gabsoftware:
Fixes webbukkit#2533

Most of the work was done by @tiagoad (commit https://github.com/tiagoad/dynmap/commit/8fc95bfc96949b6286bfc03e678407c3ebaf453f)

But a few bits had to be changed or the compilation would fail.

Signs do not disappear now.


mikeprimm:
Need to review yours vs @tiagoad 's PR - will accept whichever one looks cleaner - thanks!

HuJK:
So,this code will be in Dynmap v3.0-beta-6?

PR #2633, opened 2019-10-21
Resume full renders (skip already rendered tiles)

### LINKED ISSUE #2630: Continue Cancelled Fullrender?
I have a world that is half rendered with [ChunkyMap](https://github.com/leMaik/ChunkyMap), but I had to cancel the render on it. ChunkyMap takes a LONG time, so I would rather not start from scratch. 

Is there a way to continue the fullrender on tiles that are not yet rendered? It should be possible: before any tile is sent to be rendered, check the storage for that tile. If it exists, skip tile render.

Where in dynmap's code could I start looking for a way to implement this? Where is the tile set up for render during fullrender?

Marvosg:
Implemented functionality to resume full renders by typing "dynmap fullrender resume \<world\>" or "dynmap fullrender resume \<world\>:\<map\>". At the start of the render, existing map tiles are loaded from storage and their location info saved in a set. Before each tile is rendered, the set is checked if it contains the tile about to be rendered. If so, the tile is skipped. Information about skipped tiles is output in the periodic tile update message.


mikeprimm:
Looks promising - need to review further - thanks!

mikeprimm (approved):

PR #2668, opened 2019-11-29
Standalone: Load changes immediately after page load

lennartbecker:
This loads changes immediately after the page load without waiting for the updaterate.


pkramme:
Thanks for implementing! This is really helpful for me, because on my server the updaterate is set to 30 seconds. This means that my users have to wait 30 seconds after the pageload before they see the weather, time, and other players, which is annoying when you just want to see the current players fast.

Hopefully this will get merged!

pkramme:
I have locally patched my map.js and it seems to work as expected.

mikeprimm:
This looks good - thanks!

PR #2691, opened 2019-12-12
Added support for 1.15

655456:


mikeprimm:
Looks good - will review - there is a second PR for the same things, so will likely accept one of them after reviewing both.  Thanks!

PR #2692, opened 2019-12-12
spigot 1.15 support

pk910:
Patch for 1.15 support (see https://github.com/webbukkit/dynmap/issues/2689#issuecomment-565147139)

In minecraft 1.15 they've changed the way how biome information is stored.
There is a new class called `BiomeStorage` available that manages biome information.


mikeprimm:
Looks good - will review - there is a second PR for the same things, so will likely accept one of them after reviewing both.  Thanks!

pk910:
Yea, sure :)
Maybe a combination of both will make it.. The other implementation of the changes in BukkitVersionHelperGeneric.java looks much cleaner than mine.
See https://github.com/webbukkit/dynmap/issues/2689#issuecomment-565686719 for explanation of the second commit.

mikeprimm:
Yep - for big PRs like this, I almost always need to do some 'touches' before it's accepted.  I really appreciate the PRs, though!

PR #2696, opened 2019-12-15
fix water color

iruis:
MC 1.14 BiomeBase.m() reference from class BiomeBase.a variable h
MC 1.15 BiomeBase.o() reference from class BiomeBase.a variable h
MC 1.15 BiomeBase.p() reference from class BiomeBase.a variable i

BiomeBase.a veriable h is waterColor
BiomeBase.a variable i is waterFogColor


mikeprimm:
Already got this fixed while I was handling the other 1.15 PRs - thanks for the confirmation!

PR #2748, opened 2020-01-13
Update Jetty to 9.4.24

### LINKED ISSUE #2762: 1.15.2 snapshot breaks my dynmap.
After updating to 1.15.2 and the 1.15.2 dynmap snapshot, my dynmaps are blank. I then tried the creative server but left the window open, after updating etc I refreshed and a message at the top of the window said https://i.tws.gg/tcXApk1W9B.png Here is a pic of the dynmap https://i.tws.gg/qwUJl0sQ2o.png

When I Downgrade to 1.15.1 and the 1.15 dynmap it works fine

Ry0taK:
# Summary
This pull request updates Jetty to 9.4.24. Due to the size of the version difference, it may break various functions of dynmap. I've tested some, and confirmed it's working. However, please test it carefully.

(Tested with Spigot 1.12.2)

### DIFF
--- a/DynmapCore/build.gradle
+++ b/DynmapCore/build.gradle
@@ -2,10 +2,10 @@ description = "DynmapCore"
 
 dependencies {
     compile project(path: ':DynmapCoreAPI', configuration: 'shadow')
-		compile 'javax.servlet:javax.servlet-api:3.1'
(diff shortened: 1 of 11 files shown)


mikeprimm:
Looks good - I need to do some more testing, but thanks for the good work!!

PR #2769, opened 2020-01-27
Add Vault compatibility and fix PEX v2 usage

### LINKED ISSUE #2730: PEX v2 Support
Looks like this might be the cause of #1969 as well

PEX v2 is in development and has an API entirely different from PEX v1 -- however Dynmap's permission abstraction layer only detects PEX by the plugin name, rather than any classes that may be present. This means that Dynmap tries to call the PEX v1 api which of course no longer exists, leading to a whole load of fun stacktraces.

Unfortunately I am not at a point where I'm willing to call the PEX v2 API stable, so I don't expect you to go ahead and write a permission resolver for the new PEX version. However, there are a few other options that would at least make the situation better even if they don't give full compatibility until release:

- Check that the PEX v1 `ru.tehkode.[...].PermissionsEx` class exists, and if not just fall back to the superperms resolver. (the simplest option). This looks like it has limited compatibility for offline players though.
- Just use Vault -- I'm assuming there's a historical reason for having your own permissions handling, but keep in mind I will happily maintain my own Vault provider that I'll update for any PEX api changes :)
- Write and maintain your own PEX v2 resolver -- would give full compatibility but again there may be sudden breakages as I work towards an eventual stable release, so probably the highest maintenance burden.

What do you think? It would be great to have even partial compatibility between Dynmap and PEX2.

zml2008:
This would resolve #2730 without having to depend on PEX's internal API by:

a) Making sure that Dynmap's PEX v1 permission resolver only tries to register itself when PEX 1 is running on the server
b) Adding a Vault resolver to support PEX 2, as well as any other permissions plugin that may be created

I've also tossed in a change to how maven central is referenced, since that no longer supports plain http access -- that was mostly just needed so I could make dynmap build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,10 +8,9 @@ plugins {
 allprojects  {
   repositories {
      mavenLocal()
-     mavenCentral()
      maven { url "https://oss.sonatype.org/content/repositories/releases" }
      maven { url "http://repo.mikeprimm.com" }
-     maven { url "http://repo.maven.apache.org/maven2" }
+     maven { url "https://repo.maven.apache.org/maven2" }
      maven { url "https://hub.spigotmc.org/nexus/content/repositories/snapshots/" }
      maven { url "https://repo.codemc.org/repository/maven-public/" }
   }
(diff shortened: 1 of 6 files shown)


mikeprimm (approved):

mikeprimm:
Looks good - thanks!

PR #2781, opened 2020-02-06
Fix unintended redirection to welcome file

### LINKED ISSUE #2765: 302 redirect from "/" to "/index.html" is new behavior that can partially break some configurations
**Known affected versions:**
beta-10

**Previous (1.15.1 version) behavior (desired):**
Directly returns the index.html page even on the "/" path.

**Current undesired behavior:**
If the root path on the builtin webserver is accessed, it will return a 302 redirect to the `/index.html` path. The full URL passed in location header for the 302 response was set to `http://<requests-host-header>/index.html`.

**Notes:**
I use Cloudflare in front of the webserver. I rewrite the request so that from the user perspective they are accessing over HTTPS on my domain. On the backend it accesses http, but on a different domain. No direct access is allowed to the configured IP or port except through Cloudflare.

Technically, I think this new behavior would also break redirects when accepting traffic on the IP itself and not using port 80, since the port is not preserved in the location header.

 **Workaround:**
I manually configured Cloudflare to intercept requests to the "/" path and rewrite them to "/index.html" using the normal protocol and domain that my users are accessing.

**Possible (untested) solutions:**
1. I think it would break less situations if when it constructed the location response header, it added the configured port appended to the domain that was in the host header.

2. Alternatively, and preferred for me personally, allow the full root URL (with protocol, domain, and port) to be configured so that the config could give the webserver more information about how user's were accessing it, allowing any redirects to access using the correct protocol, domain, and port for that deployment.

3. Both. Maybe set the default behavior to be the first solution, and also allow a configured override as described in the second solution?

4. Revert to the old behavior.

**edit:** clarified after further testing on the old version.

Ry0taK:
This pull request fixes #2765 which caused by #2748. 
(It'll update Jetty to the latest version which released in this year.)
Eclipse Jetty changed the internal behavior of welcome file. I thought it can be fixed with `setRedirectWelcome(false)`, however, it wasn't working correctly.
`setRedirectWelcome` requires context to not be null. So this pull request wraps `FilterHandler` with `ContextHandler`.
I believe this won't break anything, however, it still may break something that I unexpected.
(Tested with Spigot 1.12.2)

### DIFF
--- a/DynmapCore/build.gradle
+++ b/DynmapCore/build.gradle
@@ -3,8 +3,8 @@ description = "DynmapCore"
 dependencies {
     compile project(path: ':DynmapCoreAPI', configuration: 'shadow')
     compile 'javax.servlet:javax.servlet-api:3.1'
-    compile 'org.eclipse.jetty:jetty-server:9.4.24.v20191120'
-    compile 'org.eclipse.jetty:jetty-servlet:9.4.24.v20191120'
+    compile 'org.eclipse.jetty:jetty-server:9.4.26.v20200117'
+    compile 'org.eclipse.jetty:jetty-servlet:9.4.26.v20200117'
     compile 'com.googlecode.json-simple:json-simple:1.1.1'
     compile 'org.yaml:snakeyaml:1.23'
     compile 'com.googlecode.owasp-java-html-sanitizer:owasp-java-html-sanitizer:20180219.1'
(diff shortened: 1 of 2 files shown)


mikeprimm:
Looks good - will roll into next update

mikeprimm-triva:
Dev build here - https://dynmap.us/builds/dynmap/Dynmap-3.0-SNAPSHOT-spigot.jar

PR #2839, opened 2020-03-31
fix vertical centering of markers

L0laapk3:
I've noticed that all my markers weren't vertically centered properly.
the `margin-top: -8px;` in `.dynmap .mapMarker .markerIcon16x16` wasn't actually working correctly, it was getting limited by something else beyond -3px. Changing this to `transform: translate(-50%, -50%);` fixed the issue for me.


mikeprimm (approved):
Looks good!

PR #2838, opened 2020-03-31
Support Mohist

Mgazul:

PR #2874, opened 2020-04-13
fix Bukkit version detection

0uti:
changes from PR #2838 has broken Bukkit version detection.

Log: https://hastebin.com/galideqebu.md


mikeprimm:
From what I can see, this will fix the 'normal' path, but break the path added by the other PR (since the alternative path that was added by that PR will never get reached).  Need to get both ironed out - probably by using the NoFail with a null test to hit the alternative path.

0uti:
closed this PR. 
breaking change got reverted #2875

PR #2927, opened 2020-05-11
WIP Allow using %uuid% in skin-url

AliceDTRH:
This simple change allows you to use %uuid% inside skin-url allowing us to lookup skins on newer versions of minecraft (assuming you have an api setup for it).


Tydium:
Would there be any luck on merging this?

PR #2994, opened 2020-06-27
Add required API version declaration in plugin.yml

jchoyt:
Reduces errors when started on spigot


KisaragiEffective:
@jchoyt not `1.6` but `1.16`

RobinWhitfield:
I wonder what issues came up that caused this reversion:
https://github.com/webbukkit/dynmap/commit/178681d28413aa49d42100bf689b5f889a528409

mikeprimm:
This is broken

mikeprimm:
The api-version change breaks any Spigot version older than the version you put in, and Dynmap supports versions back to before 1.10 with the same binary.

Issue #3083, opened 2020-08-12
Invalid length given for storage, got: 90 but expected: 64


ghost:
Tile generation halts after several hundred tiles.

* **Dynmap Version:** *dynmap v3.1-beta-2-389*
* **Server Version:** *Spigot 1.16.1*

```[09:37:01] [Dynmap Render Thread/INFO]: [dynmap] Loading default resource pack
[09:37:10] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 100 tiles rendered (92.35 msec/tile, 55.73 msec per render)^[[m
[09:37:18] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 200 tiles rendered (85.99 msec/tile, 53.50 msec per render)^[[m
[09:37:25] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 300 tiles rendered (78.82 msec/tile, 47.06 msec per render)^[[m
[09:37:33] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 400 tiles rendered (75.78 msec/tile, 43.85 msec per render)^[[m
[09:37:40] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 500 tiles rendered (74.18 msec/tile, 42.84 msec per render)^[[m
[09:37:46] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 600 tiles rendered (72.44 msec/tile, 41.91 msec per render)^[[m
[09:37:53] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 700 tiles rendered (71.22 msec/tile, 41.35 msec per render)^[[m
[09:37:59] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 800 tiles rendered (69.70 msec/tile, 40.54 msec per render)^[[m
[09:38:06] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 900 tiles rendered (70.30 msec/tile, 40.21 msec per render)^[[m
[09:38:15] [Dynmap Render Thread/INFO]: Full render of map 'flat' of 'world' in progress - 1000 tiles rendered (71.43 msec/tile, 40.81 msec per render)^[[m
[...]

Issue #2697, opened 2019-12-16
Interest in developing a Fabric mod for 1.15.x


magneticflux-:
I'm interested in making a Fabric version in the same style as the existing Forge mods. How should I structure it to use the shared APIs? Are they published in a maven repo somewhere that I could use?

I don't think it would be possible to include that build in this repository, but have you considered using Gradle Composite Builds instead of a monorepo with project dependencies?

mikeprimm:
Honestly, having multiple repos for years was a nightmare, and I'd definitely not go back to it.  That said, the build still produces multiple artifacts, including the common core module that is 95% of the code and data content of dynmap (DynmapCore), so that component is importable into another build.

magneticflux-:
Are the artifacts published on a Maven repo somewhere? It looks like some of them are [here](http://repo.mikeprimm.com/us/dynmap/), but that seems out-of-date.

magneticflux-:
@mikeprimm It looks like your artifact publishing configuration is completely missing right now. `publish` and `publishToMavenLocal` are always up-to-date because there are no publications defined. Are there plans to make drastic changes, and/or would you accept a PR that gets artifacts published again?

I have parts of Dynmap working with Fabric on 1.15.1, but I'm running into issues/oddities with the builds from August 2018 (18 months old now).

magneticflux-:
I have the webserver working now, but no tile rendering yet. The implementation of `DynmapServerInterface` for Fabric is bare-bones currently.

![image](https://user-images.githubusercontent.com/9124288/71628571-3c14d400-2bbe-11ea-96b8-e8288d167aff.png)

LarveyOfficial:
Any Updates on this or has it been dropped?

magneticflux-:
@LuisVervaet I haven't made much progress beyond what I showed here, but I made a repo for what I have: https://github.com/magneticflux-/dynmap-fabric

LarveyOfficial:
Awesome! I'm exited to see if this can be done!

Martmists-GH:
Update: development has moved over to [Essentials for Fabric](https://github.com/nyliummc/Essentials/Essentials)

FedUpWith-Tech:
Is the author still working on this port? The repo hasn't been updated recently

Martmists-GH:
[...]

PR #3089, opened 2020-08-14
Fabric support (1.16.1 and 1.16.2)

lhns:
See https://github.com/webbukkit/dynmap/issues/2697#issuecomment-673095879 for previous discussion.
Until this is merged you can get the unofficial version here: https://github.com/LolHens/dynmap-fabric

I have already integrated the two new submodules into the build process. Maybe the artifact paths need to be changed.
Also there are ~~3~~ 2 new TODOs in the code (see second and third commit) which should probably be adressed in the future but for now they have workarounds and they don't seem to break anything.
~~EDIT: there seems to be a bug which causes chunks to look corrupted. We probably want to wait for that to be fixed first.
https://github.com/LolHens/dynmap-fabric/issues/19~~
~~EDIT: implemented a new workaround that doesn't seem to have said issues.~~
~~EDIT: just saw #3083. That's the same issue we are having on the fabric version. The fabric version just skips the chunk and therefore doesn't crash...~~
EDIT: fixed properly

@mikeprimm should I target master with this PR or do you want to create a fabric branch so you can tinker with this and change stuff before merging it into master?


mikeprimm:
v3.0 is the right branch - this looks good.  My main thing is that I need for anyone who does PRs and the like on Dynmap - particularly anything that touches DynmapCore or DynmapCoreAPI - to build (and hopefully test, but at least build) their chances on EVERYTHING.  Back when the components were separate repos, I cannot tell you how often PRs were OK for just one platform, and broke everything else.... big reason we're not going back to anything that is a separate branch or repo, and why I'll probably do some work to get this stitched into a single build step (even if that is just a top level script that runs 'gradle' twice :)

lhns:
That should already be that way. I integrated the two new platforms into the main gradle project.

mikeprimm:
Perfect - looking forward to spending some time reviewing it after I release tonight, and we'll get things integrated!

mikeprimm (approved):

mikeprimm:
This looks really good - I've merged it, and will be putting in some time to make sure that building and such by my process works as it should, and that the resulting artifacts are useful.  I'll kick a post here with links to the resulting builds - I'd very much appreciate some 'sanity checking' of these merged and formally built SNAPSHOT builds by any of the folks that have already been working with the unofficial builds.

PR #3095, opened 2020-08-16
fix command response being sent to all players

lhns:
in 1.15 and before the response to a command is only sent to the player who executed it.
Since 1.16 it is sent to all players. I don't think this was intended and so I created this PR.


mikeprimm (approved):
Looks good - that's odd :)

PR #3115, opened 2020-09-01
[Fabric] Make it so dynmap commands can be proxied via `/execute` without failing to parse.

i509VCB:
Also adds some gitignore entries for stuff generated by fabric at runtime.


FedUpWith-Tech:
@mikeprimm this PR looks good to me and does indeed fix the /execute behavior

PR #3133, opened 2020-09-22
Fix comma on some locales when formatting float and double values

electron93:
`String.format()` respects the current locale.

I tested it on my server, the problems with markers are gone.


DylanDelobel:
This fix work for me aswell

FedUpWith-Tech (approved):

mikeprimm-triva (approved):

mikeprimm:
Good catch - really need to just kill this code (its REALLY old, and predates when I started using the json-simple library).

PR #3168, opened 2020-11-06
Improving map icon logics and correcting Surface mapname in templates.

Kimax89:
Alot more versatile for custom maps and icons.

This logic will always refer to the naming in worlds.txt or default world naming, when using the option "deftemplatesuffix" in Configuration.txt

The icon for "world_nether:flat" will be block_nether_flat.png

Then lets say you have a custom map rendering to only show the top of the nether and you named the rendering "top" in the worlds.txt under the map nether. Then the icon will be block_nether_top.png

### DIFF
--- a/DynmapCore/src/main/resources/extracted/web/js/map.js
+++ b/DynmapCore/src/main/resources/extracted/web/js/map.js
@@ -280,18 +280,19 @@ DynMap.prototype = {
 
 				var worldName = wname;
 				var mapName = mapindex;
-				if (worldName.endsWith('_nether') || (worldName == 'DIM-1')) {
-				   worldName = 'nether';
-				   mapName = (mapindex == 'nether') ? 'surface' : 'flat';
+				//Standart minecraft world names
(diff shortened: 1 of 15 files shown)


DynmapCore/src/main/resources/extracted/web/js/map.js
```diff
@@ -280,18 +280,19 @@ DynMap.prototype = {
 
 				var worldName = wname;
 				var mapName = mapindex;
-				if (worldName.endsWith('_nether') || (worldName == 'DIM-1')) {
-				   worldName = 'nether';
-				   mapName = (mapindex == 'nether') ? 'surface' : 'flat';
+				//Standart minecraft world names
+				if(worldName.startsWith('world_')) {
```
mikeprimm, line 284: This doesn;'t work for non-default world names

mikeprimm:
Sorry for delayed review, but making assumptions like "world_" as a standard world name is too brittle, as many folks change the default world name.   Cannot integrate this as proposed.

PR #3179, opened 2020-11-12
Fix: #3171 (StringIndexOutOfBoundsException)

### LINKED ISSUE #3171: String index out of range: 6 error, causes the dynmap to stop updating.
**Issue Description:** Randomly Dynmap throws an error, causing the map to stop update/players frozen in the map.

* **Dynmap Version:** v3.1-SNAPSHOT-429
* **Server Version:** Tuinity 1.16.4
* **Pastebin of Configuration.txt:** *https://gist.github.com/MrPowerGamerBR/9b8f62bace39b8636096862f4f59d569*
* **Server Host (if applicable):** OVH
* **Pastebin of crashlogs or other relevant logs:** *https://gist.github.com/MrPowerGamerBR/b7e66abb461bdac628fe6e87660c9073*
* **Other Relevant Data/Screenshots:** -
* **Steps to Replicate:** To be honest I'm not able to replicate this yet aside that "this happens randomly", is there's any way to enable debug logs to make it easier to try to track this issue?

The only thing that I know is that, after this error happens, dynmap stops working (you can still see the map but it won't update, players are also frozen in the map)

 [X] *I have looked at all other issues and this is not a duplicate*  
 [ ] *I have been able to replicate this*

If I had to guess I think it is related to strings with latin characters (my player base is from Brazil), so when dynmap tries to strip colors from the string (?), it throws an error.

If I had to guess *by looking at the source code*
```java
    // Strip color - assume we're returning safe html text
    public static String stripColor(String s) {
        s = DynmapChatColor.stripColor(s);    /* Strip standard color encoding */
        /* Handle Essentials nickname encoding too */
        int idx = 0;
        while((idx = s.indexOf('&', idx)) >= 0) {
            char c = s.charAt(idx+1);   /* Get next character */
            if(c == '&') {  /* Another ampersand */
                s = s.substring(0, idx) + s.substring(idx+1);
            }
            else {
                s = s.substring(0, idx) + s.substring(idx+2);
            }
            idx++;
        }
        // Apply sanitize policy before returning
        return sanitizeHTML(s);
    }
```

I didn't test, but what would happen if dynmap tried to strip color from a "hello&" string? Wouldn't it cause a IndexOutOfBoundsException since it is always trying to get the `idx + 1` character?

WouterGritter:


mikeprimm (approved):

mikeprimm:
Thanks!

WouterGritter:
Was just looking through this old pr of mine, as normal people do right?
This actually never made it into the v3.0 branch, I guess the target was set wrong.

mikeprimm-mirific:
Yep - 'recommended' is an archive branch (that is, I push releases TO it, but nothing ever comes FROM it).

PR #3184, opened 2020-11-15
Null check for path environment variable.

### LINKED ISSUE #3170: findExecutableOnPath, for the new webp feature, can cause NPE in some environments
**Issue Description:** 

Commit 460e6f981 of 2020-10-25 adds a new method `findExecutableOnPath` that calls `System.getenv("PATH").split ...`. In a Multicraft environment, java seems to get called without `PATH` being defined, probably for security reasons. This results in `System.getenv("PATH")` returning `null`, and the method throwing a NPE, preventing dynmap initialization.

Solution: make the method return null if `System.getenv("PATH")` returns null.

 [X] *I have looked at all other issues and this is not a duplicate*  
 [X] *I have been able to replicate this*

silverwolfg11:
Some systems do not set a path environment variable for whatever reason and thus `System.getenv("PATH")` returns a null value for the string. 

Closes #3170

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
@@ -438,7 +438,12 @@ public class DynmapCore implements DynmapCommonAPI {
     }
 
     private String findExecutableOnPath(String fname) {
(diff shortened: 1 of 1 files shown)


mikeprimm (approved):

mikeprimm:
Thanks!

PR #3196, opened 2020-11-24
Fix fullrender deleting the files in the new format instead of the old

### LINKED ISSUE #3195: fullrender deletes the wrong images when rendering finishes
**Issue Description:**
~~After switching from jpeg to WebP, the WebP images get rendered and appear on the filesystem, but when rendering finishes, the WebP files get deleted, while the old jpeg files remain.~~
I originally assumed this was an issue with WebP specifically, which is not the case.
When switching from one file format to another, and then doing a fullrender, the tiles in the new format get generated properly and appear on the filesystem, but when rendering finishes, all new tiles get deleted.
I assume what fullrender is trying do here is to delete old tiles if there are any, but it instead deletes the new tiles. This behavior only affects the fullrender command, when doing a radius render, the new tiles stay.
I tested this by switching from JPEG to WebP, and, after removing the tiles, by switching from WebP to JPEG. I observed the above behavior both times.

* **Dynmap Version:** 3.1-beta5-spigot
* **Server Version:** Spigot git-Spigot-a19903d-167ff59 (MC: 1.16.4)
* **Pastebin of Configuration.txt:** https://pastebin.com/6ArKTxRv
* **Server Host (if applicable):** local
* **Pastebin of crashlogs or other relevant logs:** 
* **Steps to Replicate:**
  * Set up a fresh and clean Spigot with Dynmap, don't change Dynmap settings
  * Start the server, run `dynmap fullrender world`, wait till it finishes, stop the server
  * Change the `image-format` setting to  `webp-q100`
  * Start the server, run `dynmap fullrender world`, and watch `plugins/dynmap/web/tiles/world/t/-1_-1/`.

  As a faster alternative, you can run `dynamp fullrender world:flat` and watch the respective directory.

 - [x] *I have looked at all other issues and this is not a duplicate*  
 - [x] *I have been able to replicate this*

thrimbor:
WebP support was recently added, but only PNG and JPG were hardcoded in FileTreeMapStorage. This caused an issue with running fullrender after changing the image format in the config file, which tried to delete the outdated files using the *current* image format.
This removes the hardcoded values and replaces them with code that uses all the ImageEncoding enum values except the one matching the current format.

Fixes #3195.

There's still a remaining issue with the zoom tiles not getting regenerated after an image format change, but I don't want to dig into that tonight.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/storage/filetree/FileTreeMapStorage.java
+++ b/DynmapCore/src/main/java/org/dynmap/storage/filetree/FileTreeMapStorage.java
@@ -73,15 +73,17 @@ public class FileTreeMapStorage extends MapStorage {
             }
             return ff;
         }
-        private File getTileFileAltFormat() {
+        private List<File> getTileFilesAltFormats() {
             ImageEncoding fmt = map.getImageFormat().getEncoding();
-            if (fmt == ImageEncoding.PNG) {
-                fmt = ImageEncoding.JPG;
+
+            List<File> files = new ArrayList<File>();
+            for (ImageEncoding ie: ImageEncoding.values()) {
+                if (ie != fmt) {
+                    files.add(getTileFile(ie));
+                }
             }
-            else {
-                fmt = ImageEncoding.PNG;
-            }
-            return getTileFile(fmt);
+
+            return files;
         }
         @Override
         public boolean exists() {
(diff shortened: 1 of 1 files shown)


mikeprimm (approved):

mikeprimm:
Thanks!!

PR #3223, opened 2020-12-26
Allow for URI-unsafe player names

mooinglemur:
Some plugins allow for a player name prefix. If this contains a username-unsafe character like %, it can break player face rendering on the map.


mikeprimm (approved):

mikeprimm:
Looks good - thanks!

Issue #2345, opened 2018-09-16
Support for Async Chunk Loading (Paper)


aikar:
Hi mike, it's been a long time since we spoke!

Paper has an API addition that allows plugins to request chunks asynchronously. This API has existed since 1.9 roughly.

I'm requesting DynMap add support for this by detecting if World#getChunkAtAsync exists, and if it does, use an alternative chunk loader that uses the Paper API method.

By doing this, Paper will be able to asynchronously load all chunks. We also have a World#isGenerated(x,z) method that you may find useful too since you were using NMS for that logic already.

I am also planning to add a new method in 1.13 to "getChunkAtAsyncWithoutLoading" kind of concept, that would suit dynmaps needs, in that we can load the chunk but never post it into the world.

This would remove your need to immediately unload the chunk.

Many server owners are switching to Paper, and I'm hoping you can add this support so that dynmap can be nearly free in performance cost.

I would love if you would join the Paper discord, and I can speak more with you on there if you need :) https://paperdiscord.emc.gs

Thanks, hope to reconnect soon!

smmmadden:
Hi Aikar, as mentioned in Discord not sure if you saw this or not or if it's something Mike would need to make changes for, but wanted to share nonetheless.  https://github.com/webbukkit/dynmap/issues/2344

FWIW - I don't know whether my timings are typical or atypical but it is at least one benchmark. -Steve

jacklollz2:
I'd love to see Async chunks implemented! The server performance runs very poorly when dynmap generates chunks.

generrosity:
I'm going to out on a limb, but my understanding from talking to the only dev is that 'Spigot' and 'Forge' are the two primary platforms that he works on and support, and SpongeForge and Paper do get consideration but no active work. There isn't a Paper fork, so comiling a feature not in Spigot would likely throw errors on other platforms.

I'm sure multithreading is on the back of his mind as an optimisation, but at this stage he is likely focused purely on 1.13 implementation.

If you are interested in the code, I think passive chunk updates are done as on-thread events, and the /dynmap fullrender is a distinct command but could be easiest split off if there was a "easy" way to a) confirm we were running Paper not Spigot and b) could run the async execution without upsetting Spigot for just having the command nearby

I hope that makes sense?

smmmadden:
[...]

PR #3281, opened 2021-02-13
PaperMC - Asynchronous Chunk Loading Support

crystall1nedev:
This draft pull request is for my work on getting the multithreaded rendering support available for Dynmap with servers running PaperMC, for issue #2345. Here's what I want to get done:

- [x] Attach PaperLib to Spigot's build.gradle
- [ ] Define new classes and separate Spigot classes 
- [ ] Call both sets of classes so backwards compatibility with Spigot is possible
- [ ] Set Dynmap to detect whether or not Paper is in use

Having Dynmap capable of asynchronous chunk loading on PaperMC would present a major performance improvement on servers with the software.

Some of my commits deal with a file called `.travis.yml`. This file is so that I can run `./gradlew build` to test my code changes with a better computer than I have available for compilation. This file won't be on my final pull request.


Doregon:
This looks like it needs a major rewrite of Dynmap to work, and I thought that the experience I had with Java would be good. Sadly, I cannot continue because I need more training with Java. I may continue this in the future but for now I cannot.

Doregon:
Alright, I forgot to write the reason why I closed this: I was swamped over the past couple of weeks and had to work harder on my own personal Java skills, but I'm installing IntelliJ IDEA right now and I'll start up with it again. I refuse to give up until we reach async functionality in Dynmap! I'll be opening a new pull request, though, because I need to reorganize and regroup.

PR #3338, opened 2021-04-12
Add option to disable HTML Sanitizer

### LINKED ISSUE #2288: [Collection] Text formatting issues in Marker names
Putting single quotes in marker names are being displayed as the HTML character `&#39;`  
  
Example:  
`recon's house` is being displayed as `recon&#39;s house` on the live map

Joo200:
I added the option to disable the HTML Sanitizer as I have some issues with changing html snippets via the api.
The default behavior is unchanged as the default value is set to "true".

Maybe this should include a check whether the label is markup or not.

It should close https://github.com/webbukkit/dynmap/issues/2288

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
@@ -122,6 +122,7 @@ public class DynmapCore implements DynmapCommonAPI {
     private boolean transparentLeaves = true;
     private List<String> sortPermissionNodes;
     private int perTickLimit = 50;   // 50 ms
+    private boolean sanitizeHtml = true;
     private boolean dumpMissing = false;
     private static boolean migrate_chunks = false;
     public boolean isInternalWebServerDisabled = false;
@@ -534,6 +535,8 @@ public class DynmapCore implements DynmapCommonAPI {
         updateplayerlimit = configuration.getInteger("updateplayerlimit", 0);
         /* Load sort permission nodes */
         sortPermissionNodes = configuration.getStrings("player-sort-permission-nodes", null);
+        /* Whether to sanitize html labels and descriptions */
+        sanitizeHtml = configuration.getBoolean("sanitize-html", true);
         
         perTickLimit = configuration.getInteger("per-tick-time-limit", 50);
         if (perTickLimit < 5) perTickLimit = 5;
@@ -2534,6 +2537,7 @@ public class DynmapCore implements DynmapCommonAPI {
     public int getMaxTickUseMS() {
         return perTickLimit;
     }
+    public boolean shouldSanitizeHtml() { return sanitizeHtml; }
     public boolean dumpMissingBlocks() {
         return dumpMissing;
     }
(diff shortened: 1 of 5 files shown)


Joo200:
@mikeprimm  - can you give any update here? I don't want to maintain my own fork for that config option.

Thanks

ghost:
I am also facing some problems that are caused by the HTML sanitizer and so i want to disable it.

May somebody can take a look at this PR?

mikeprimm-triva:
Turning off the HTML sanitizer is a security concern - it allows a misbehaving user and/or plugin to inject hostile content into a viewers web browser (e.g <script> tags), so this is not a particularly attractive PR, IMHO.   I received a formal CERT (security defect) advisory due to the lack of the sanitizing years back...

Joo200:
Unfortunatly you can bypass that sanitizer by using the api to update the marker's description. (open the map, update the marker after that, unsanitized html. You just need a plugin which will update the marker every 10 seconds)
I understand your security concern but every admin can create misbehaving stuff and the new config option is disabled by default.

I need that option to disable removing class tags, e.g. from
```xml
<img class="markerIcon16x16" src="tiles/_markers_/building.png">
<span class="markerName markerName_survivalCities markerName16x16">
<div class="cityHoverBox" data-city-id="38">
<div class="name">Jatorri</div>
<div>Stufe: Stadt</div>
<div>Freie Grundstücke: 16/47</div>
</div>
</span>
```
grafik

it will transform to
```xml
<img class="markerIcon16x16" src="tiles/_markers_/building.png">
<span class="markerName markerName_survivalCities markerName16x16">
<div>
<div>Jatorri</div>
<div>Stufe: Stadt</div>
<div>Freie Grundstücke: 16/47</div>
</div>
</span>
```
grafik


The sanitizer will remove currently the data-city-id which we use to add a clickable text with javascript (modified index.html, added a custom js module). You can see an example here: https://esconia.map.terraconia.de/
The marker icons are clickable but currently I have to use the version from my this branch.

mikeprimm-triva:
This isn't about server admins - its about the fact that users on a server can create and control text in markers, whether via mods that allow name aliases or via adding sign-based markers or the like.   Marker text is user-accessible content, not just server admin level content.   A hoster that controls the mods on the server needs to be trusted, but not necessarily the folks with op roles (much less lower).

Allowing the sanitizer to NOT strip class attributes is probably OK - that's aesthetic impact, not XSS or script injection.  I'd be willing to provide a setting to turn that off.

Joo200:
But not every server has those sign-based or name aliases enabled. I don't have users with such permission nor do I have sign-based or name aliases. I use that modified informations for the link to the information page from the city and it's plugin written.
And yes, not everyone is trusted but there's a reason I added it as a config option.

Unless you add some list with "good attributes" to the config I don't see how I should have my features with the sanitizer as I need the data-city-id as attribute for the link.

Maybe it's easier to maintain my own fork...

mikeprimm:
Maintaining your own fork is best - I'm not allowing the sanitize to be turned off in general - I've got 20,000 servers worth of community, and many in hosted environments, and I'm not reintroducing an injection attack vector that I've already been cited for publicly.  I am willing to dial it down to allow functionally safe content, like class attributes, which only present presentation concerns.

PR #3396, opened 2021-06-20
Fix SkinsRestorer API integration

### LINKED ISSUE #3394: SkinsRestorer support is broken as of SkinsRestorer v14.1.0
**Issue Description:** Support for SkinsRestorer breaks with SR v14.1.0 because of a change in the `net.skinsrestorer.bukkit.SkinsRestorer.getSkinsRestorerBukkitAPI()` method.

* **Dynmap Version:** 3.2-beta1 (also happens with 3.1)
* **Server Version:** Paper 1.16.5 build #778
* **Pastebin of Configuration.txt:** https://gist.github.com/kdkasad/0e82ebae3a158be5efe4be9d6976db41
* **Server Host (if applicable):** N/A; self-hosted
* **Pastebin of crashlogs or other relevant logs:** https://gist.github.com/kdkasad/4167c83c489355e712029cc6081090bb

### Steps to Replicate

1. Install SkinsRestorer v14.1.0
2. Set `skinsrestorer-integration` to `true` in Dynmap's configuration.
3. Start the server with Dynmap and SR installed

 [ x ] *I have looked at all other issues and this is not a duplicate*  
 [ x ] *I have been able to replicate this*

### More info

I don't know anything about how SkinsRestorer's API works or how to use it, but since I saw that the problem was a missing method, I tried to find out where that problem was introduced. It was in SkinsRestorer commit [c454a09e28e2de94adc4986605eb36398a4dee04](https://github.com/SkinsRestorer/SkinsRestorerX/commit/c454a09e28e2de94adc4986605eb36398a4dee04).

`SkinsRestorer.skinsRestorerBukkitAPI` was changed to `SkinsRestorer.skinsRestorerAPI`, which does not have a getter, and therefore cannot be accessed from Dynmap.

I'm sure that that is the problem, however I don't know the proper way to access the SkinsRestorer API since this change has been made.

kdkasad:
Fixes #3394

The dynmap plugin fetches the `SkinsRestorerAPI` instance using an unsupported method that has been removed in SkinsRestorer v14.1.0. This PR changes that to use the method suggested by the SkinsRestorer devs. It also updates the SR dependency to the latest version See the linked issue for more details.

I've tested this change with SR v14.0.2 and SR v14.1.0 and it works as expected with both versions of SkinsRestorer.

Also SR v14.1.0 will need to be supported at some point because it's the only release of SR that supports MC v1.17.

### DIFF
--- a/spigot/build.gradle
+++ b/spigot/build.gradle
@@ -15,7 +15,7 @@ dependencies {
     compile 'me.lucko.luckperms:luckperms-api:4.3'
     compile 'net.luckperms:api:5.0'
     compile('com.github.MilkBowl:VaultAPI:1.7') { transitive = false }
-    compileOnly 'net.skinsrestorer:skinsrestorer:14.0.0-SNAPSHOT@jar'
+    compileOnly 'net.skinsrestorer:skinsrestorer:14.1.0-SNAPSHOT@jar'
     compile project(":dynmap-api")
     compile project(path: ":DynmapCore", configuration: "shadow")
     compile group: 'ru.tehkode', name: 'PermissionsEx', version: '1.19.1'
--- a/spigot/src/main/java/org/dynmap/bukkit/DynmapPlugin.java
+++ b/spigot/src/main/java/org/dynmap/bukkit/DynmapPlugin.java
@@ -917,10 +917,10 @@ public class DynmapPlugin extends JavaPlugin implements DynmapAPI {
                 if (skinsRestorer == null) {
                     Log.warning("SkinsRestorer integration can't be enabled because SkinsRestorer not installed");
                 } else {
-                    skinUrlProvider = new SkinsRestorerSkinUrlProvider(skinsRestorer);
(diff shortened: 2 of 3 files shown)


mikeprimm:
Thanks - I'll probably do some more on this, as the existing exception catching is inadequate (as we saw with this error - not having the class is one thing, but undefined method is also relevant).

kdkasad:
> Thanks - I'll probably do some more on this, as the existing exception catching is inadequate (as we saw with this error - not having the class is one thing, but undefined method is also relevant).

I think it might make more sense to just check if the SR plugin is loaded using `getServer().getPluginManager().isPluginEnabled("SkinsRestorer")` since it's no longer necessary to get the `SkinsRestorer` object.

mikeprimm:
Well, the lack of a reliable binding is the problem - the plugin was called SkinsRestorer before the latest break, so doing that check would not avoid this exception - the plugin name is insufficient, without a long term stable API to depend on, to avoid throwing an unhandled exception.

PR #3411, opened 2021-06-27
Gradle 7.1

kosma:
List of changes:
* Gradle - updated to 7.1 (for Java 16 support)
* build.gradle - use new keywords (`implementation` instead of `compile`)
* build.gradle - fix many duplicated resources (error on Gradle 7)
* Loom - updated to 0.6-SNAPSHOT (minimum version that supports Gradle 7, we can possibly go higher)
* ForgeGradle - updated to 5.x (minimum version that supports Gradle 7)
* Restored Forge builds (except 1.11.2 and 1.12.2 - see below)

List of things broken:
* Forge builds of 1.11.2 and 1.12.2. The reason is that they require ForgeGradle 2, but Gradle 7 requires ForgeGradle 5. I don't have a good solution here. Perhaps using a legacy Java+Gradle+ForgeGradle version just for those? It might be necessary anyway with the fast progress imposed by Fabric people.

Caveats:
* I've not tested the builds - only verified that they build properly.


mikeprimm (approved):
Looks good!

mikeprimm:
Thanks for this - looks great, and it gets most of our Forge build issues sorted out!

PR #3412, opened 2021-06-27
Fabric 1.17

kosma:
Fabric 1.17 port, based on 1.16.4 port. Didn't require significant changes besides `migrateMappings` and a few manual fixes to code and mixins. This version has been running at my private SMP for the last 2 weeks, we've had zero problems. I consider it okay for non-critical deployments.

Important: this PR is built on top of of #3411.


mikeprimm:
Merged in, but having problems with getting it to build - I've temporarily commented it out in the base settings.gradle until we get it sorted out

PR #3428, opened 2021-07-14
Fix possible nullref

Bond-009:
This one builds `BUILD SUCCESSFUL in 2m 36s`


mikeprimm:
Handled in a different fix - thanks, though!

Issue #3418, opened 2021-07-03
Could not pass event BlockPhysicsEvent to dynmap v3.2-beta-2-483


BlackAsLight:
> template is **bold**
> sample data is *italicized*

**Issue Description:** *I get spammed with this error in the console every second some player is online.*

* **Dynmap Version:** *v3.2 beta 2 - spigot*
* **Server Version:** *paper 1.17 build 69*
* **Pastebin of Configuration.txt:** *https://pastebin.com/qd6DFzLY*
* **Pastebin of crashlogs or other relevant logs:** *https://pastebin.com/nvzV67An*
* **Steps to Replicate:** *Somebody logs onto my server and it starts spamming this until they log off.*

 [TICK] *I have looked at all other issues and this is not a duplicate*  
 [TICK] *I have been able to replicate this*

bbayu123:
I would like to add to this issue since I'm getting the same error, but the replication is different.
This time, on **Paper 1.16.5 build 779**, all the errors show that it threw the stacktrace in BlockPhysicsEvents.
Stacktrace 1: https://pastebin.com/TEHdU1r5
Stacktrace 2: https://pastebin.com/UX8Lb1us

wioxjk:
Getting this also but on 1.17.
The map works generally.

I do assume that a new .jar built for 1.17 will solve this.

Compithegreat:
Got this as well in 1.17.1.

FedUpWith-Tech:
Can replicate in Paper 1.17.1-91
With Dynmap 3.2-beta-2-483

wioxjk:
Can confirm that this is fixed in the latest version that I pulled from github and built myself

mikeprimm-triva:
Fix is in latest development builds - http://dynmap.us/builds/dynmap/Dynmap-3.2-SNAPSHOT-spigot.jar

BrainStone:
Interesting. There have been no changes between beta-2 and the latest state on GitHub.  
Unless you were referring to my PR that is.

RaidersPlay:
Tried the http://dynmap.us/builds/dynmap/Dynmap-3.2-SNAPSHOT-spigot.jar build
Seems like the exact same as the Dynmap v3.2-beta-2 that was on my server. Spammed messages still happening. 

[...]

PR #3431, opened 2021-07-18
Fix issue with getting the type id of a block

BrainStone:
This is a fix for #3418. The idea is to handle the case of when the the numeric IDs have been removed completely.

Includes some performance optimizations. Also handles the case of new materials not even having an ID.

This is really just a dirty solution.  
Ideally `DynmapServerInterface#getBlockIDAt` gets removed and all comparisons on the Spigot side get done with the Material enum. Removing this method completely should be fine, as it only seems to be used by Forge 1.11 and 1.12 when checking if something is a sign.  
Should that bigger change get approved by the project leads, feel free to reject this change in favor of that removal.

Change has been tested on 1.16.5 and 1.17.1


mikeprimm:
Looks good - thanks

Issue #3427, opened 2021-07-14
LOD map render tiles gets corrupted/does not generate after a restart


VL4DST3R:
> template is **bold**
> sample data is *italicized*

**Issue Description:** *I noticed a recurring issue lately on my map renders where if something interrupts the generation, like a server restart, the render resumes on the highest zom level (say, LOD0), it does not propagate properly to lower-res LOD's (when you zoom out).*

*This results in a portion of black lines from where the rendering was interrupted, unless you zoom in all the way. This has happened on multiple maps, across the last 2 dynmap versions. While i would obviously like to know why is this happening or if there is a fix, my main gripe is the fact that if i do a `dynmap radiusrender` for the affected areas, it doesn't even consider refreshing those non-LOD0 tiles, so essentially they can never be regenerated unless i purge the entire map and start over.*

* **Dynmap Version:** *Dynmap version: core=3.2-beta-2-483, plugin=3.2-beta-2-483*
* **Server Version:** *Paper version git-Paper-642 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)
* **Pastebin of Configuration.txt:** *https://pastebin.com/Jds2H9rP*
* **Server Host (if applicable):** *self-hosted*
* **Pastebin of crashlogs or other relevant logs:** *n/a*
* **Other Relevant Data/Screenshots:** *This happens regardless of map version and while using no texture pack*
* **Steps to Replicate:** *(for simplicity) purge a world and force a fullrender. Stop the server normally midway through rendering, then start again and let it finish. Result: (continued below)
![Screenshot_2021-07-14 Now with even more, updated nuts ](https://user-images.githubusercontent.com/32654378/125592462-bf52c3bb-d59d-4a66-8b74-9511c2a620ea.png)*

 [x] *I **have** looked at all other issues and this is indeed a duplicate of #2485! but i'm posting this since it was never an issue on our server before and wanted to know if this is*
- *something new in the last 2 dynmap versions,*
- *if it is related to my config and the use of cwebp or dwebp,*
- *OR related to using a particular paper version, since i saw it mentioned in a few threads.*  

 [ x] *I have been able to replicate this*

BrainStone:
I can confirm this.  
For the record this is called the zoom out processing/rendering. And it's incredibly broken in the current version. For me the rendering doesn't even need to be interrupted for this issue to appear but it certainly amplifies this issue.

Rerendering these regions doesn't fix it.

[...]

Issue #2485, opened 2019-05-21
[Collection] Fullrender not rendering all chunks on pregenerated non-world painter maps


CannonBoomBoom:
![image](https://user-images.githubusercontent.com/22482596/58111103-70e79400-7bb6-11e9-8a93-b0ee0255c32f.png)
I have pre-rendered the whole 10k x 10k map, and I have let the "fullrender" command run for the past couple of days. However, I am still seeing un-rendered chunks.

Version is off of your FTP: "Dynmap-3.0-SNAPSHOT-spigot.jar | 2019-05-18 22:31 | 5.4M"

MonstahGames:
I have this issue too, but for me it's random tiles scattered around the map, for you it seems to have a pattern.
Strange!

HarryFry:
I had this issue as well. I seemed to be caused by letting dynmap run while using WorldBorder to fill in the world. I solved it by first filling the world with WorldBorder and then installing dynmap and starting the full render.

ph1lt0r:
I am having the same exact issue with this, only it's on the isometric map, not the top down map. Not using world border or anything like that.

mikeprimm:
Dynmap doesn't generate chunks, and can only render those that have been properly generated.   Unless you can confirm by visiting the chunks with a real player, versus assuming the WB version you are using is working right, and then render and confirm that the known-visited chunks aren't rendering, you're probably chasing a WB and/or spigot problem.

mikeprimm:
This probably will not help, but the latest dev build (208 or later) at https://dynmap.us/builds/dynmap/Dynmap-3.0-SNAPSHOT-spigot.jar fixes as issue where chunks that were not fully generated in 1.14.x could be rendered (sometimes resulting in seeing chunks without trees or with incomplete lighting data, for example).  For what you guys are seeing, this will not necessarily be better - could be viewed as 'worse' - but it is all indicative that there might be issues with the state of chunks being driven to be generated by WB.  As I said above, visit chunks that you think are in error with a player (and run 'dynmap radiusrender N' when you get there), and see what happens: if the chunk renders properly, the state of the chunk was at issue and your player being present drove it to finish generating properly (and the chunk was either never generated, or was not completely/properly generated by WB and/or spigot) - dynmap cannot fix this, it can only show you the problem exists.

CannonBoomBoom:
Thanks Mike for looking into this.  I will test it out once I am home from work and have time!

TNTUP:
[...]

PR #3436, opened 2021-07-20
Fix zoom out processing

BrainStone:
Fixing zoom out processing, like mentioned in #3427,  #2485 and many more.

The idea is to prevent the racecondition by remembering when we started rendering and using that timestamp to save the image.

Tests are still pending.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
@@ -155,11 +155,12 @@ public abstract class DynmapWorld {
             }
             try {
                 MapStorageTile.TileRead tr = tile1.read();
-                mostRecentTimestamp = Math.max(mostRecentTimestamp, tr.lastModified);
                 if (tr != null) {
                     BufferedImage im = null;
                     try {
                         im = ImageIOManager.imageIODecode(tr);
+                        // Only consider the timestamp when the tile exists and isn't broken
+                        mostRecentTimestamp = Math.max(mostRecentTimestamp, tr.lastModified);
                     } catch (IOException iox) {
                         // Broken file - zap it
                         tile1.delete();
--- a/DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
@@ -128,6 +128,7 @@ public abstract class DynmapWorld {
     private static final int[] stepseq = { 3, 1, 2, 0 };
 
     private void processZoomFile(MapTypeState mts, MapStorageTile tile, boolean firstVariant) {
+        final long startTimestamp = System.currentTimeMillis();
         int step = 1 << tile.zoom;
         MapStorageTile ztile = tile.getZoomOutTile();
         int width = 128, height = 128;
@@ -238,7 +239,7 @@ public abstract class DynmapWorld {
                 }
             }
             else /* if (!ztile.matchesHashCode(crc)) */ {
-                ztile.write(crc, zIm);
+                ztile.write(crc, zIm, startTimestamp);
                 MapManager.mapman.pushUpdate(this, new Client.Tile(ztile.getURI()));
                 enqueueZoomOutUpdate(ztile);
             }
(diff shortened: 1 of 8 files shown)


DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
```diff
@@ -128,6 +128,7 @@ public void activateZoomOutFreshen() {
     private static final int[] stepseq = { 3, 1, 2, 0 };
 
     private void processZoomFile(MapTypeState mts, MapStorageTile tile, boolean firstVariant) {
+        final long startTimestamp = System.currentTimeMillis();
```
Pingger: The timestamp should not be the current time, but instead the time of the latest tile used within this zoomout tile

DynmapCore/src/main/java/org/dynmap/hdmap/IsoHDPerspective.java
```diff
@@ -1178,6 +1178,7 @@ public IsoHDPerspective(DynmapCore core, ConfigurationNode configuration) {
 
     @Override
     public boolean render(MapChunkCache cache, HDMapTile tile, String mapname) {
+        final long startTimestamp = System.currentTimeMillis();
```
Pingger, line 1181: The timestamp should not be the current time, but instead the time of the latest tile used within this zoomout tile

Pingger (changes requested):

DynmapCore/src/main/java/org/dynmap/hdmap/IsoHDPerspective.java
BrainStone: This is the rendering code for the lowest level of images

DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
BrainStone: Well spotted!

DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
```diff
@@ -154,6 +155,7 @@ private void processZoomFile(MapTypeState mts, MapStorageTile tile, boolean firs
             }
             try {
                 MapStorageTile.TileRead tr = tile1.read();
+                mostRecentTimestamp = Math.max(mostRecentTimestamp, tr.lastModified);
```
BrainStone: Now getting this exception:
```
[11:05:10 ERROR]: [dynmap] Exception during render job: org.dynmap.MapManager$DoZoomOutProcessing@677cad04
[11:05:10 WARN]: java.lang.NullPointerException: Cannot read field "lastModified" because "tr" is null
[11:05:10 WARN]:        at org.dynmap.DynmapWorld.processZoomFile(DynmapWorld.java:158)
[11:05:10 WARN]:        at org.dynmap.DynmapWorld.freshenZoomOutFiles(DynmapWorld.java:114)
[11:05:10 WARN]:        at org.dynmap.MapManager$DoZoomOutProcessing.run(MapManager.java:940)
[11:05:10 WARN]:        at org.dynmap.MapManager$DynmapScheduledThreadPoolExecutor$2.run(MapManager.java:245)
[11:05:10 WARN]:        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
[11:05:10 WARN]:        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[11:05:10 WARN]:        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
[11:05:10 WARN]:        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
[11:05:10 WARN]:        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
[11:05:10 WARN]:        at java.base/java.lang.Thread.run(Thread.java:831)
```

Pingger: Not all subtiles necessarily exist ...
You could do `Math.max(mostRecentTimestamp, tr==null?0:tr.lastModified);`

BrainStone: I know how to fix it ;)  
Just wanted to note it down when I saw it, because that's all that I could do in that moment.

mikeprimm:
Looks promising - thanks!

PR #3443, opened 2021-07-29
Parse Essentials RGB Hexcode nickname colors

JeffP07:
Essentials nicknames using RGB hex values end up in the format `§x§R§R§G§G§B§B`. For example, a hex color code of #1A2B3C will be formatted as `§x§1§A§2§B§3§C`.
This PR checks for the §x and then builds the hexcode from the next 6 values behind § to properly display the nickname

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/Client.java
+++ b/DynmapCore/src/main/java/org/dynmap/Client.java
@@ -226,6 +226,26 @@ public class Client {
                         break;
                     }
                 }
+                else if (c == 'x') { // Essentials nickname hexcode format
+                    if (i + 12 <= cnt){ // Check if string is at least long enough to be valid hexcode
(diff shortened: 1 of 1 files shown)


mikeprimm (approved):

mikeprimm:
Looks promising - let's give it a go.  Thanks!

mikeprimm-triva:
PR reverted - code, as submitted does not compile, so assuming it is also not tested.

JeffP07:
> PR reverted - code, as submitted does not compile, so assuming it is also not tested.

I've been using the modified version on a server since I submitted the PR, so I'm not sure what part of it isn't compiling for you

JeffP07:
Oh, I think I pushed the wrong code for the PR, will  fix and resubmit a PR later

PR #3464, opened 2021-08-22
Add Essentials RGB hex parsing

JeffP07:
Essentials nicknames using RGB hex values end up in the format §x§R§R§G§G§B§B. For example, a hex color code of #1A2B3C will be formatted as §x§1§A§2§B§3§C.
This PR checks for the §x and then builds the hexcode from the next 6 values behind § to properly display the nickname.

(This is a resubmit for #3443 to fix a mistake I made where I didn't push the correct version of the code and had misplaced my code in the wrong set of brackets)


mikeprimm:
Cool - I'll give this a look - thanks for fixing it

Issue #2564, opened 2019-07-28
Weird interaction with multiple maps for a world


stormboomer:
I've found a weird interaction when using 2 Maps for one world. 
When doing a (radius)render the world first renders the chunky map, using chunky. After the Chunky Map render is done, the "Surface Day" render is started, it then overwrites the rendered Chunky scene with the default dynmap rendering output.
This is my worlds.txt
```YML
worlds:
  - name: world
    maps:
      - class: de.lemaik.chunkymap.dynmap.ChunkyMap
        name: chunky
        title: World
        prefix: chunky
        perspective: iso_SE_30_hires
        shader: customshader
        samplesPerPixel: 10
        chunkyThreads: 8
        chunkPadding: 2
        texturepack: texturepacks/texturepack.zip
      - class: org.dynmap.hdmap.HDMap
        name: surface
        title: Surface Day
        icon: images/block_surface.png
        prefix: surface
        perspective: iso_SE_30_hires
        shader: customshader
        lighting: shadows
```
Server: git-Spigot-1a3504a-84f3da3 (MC: 1.13.2) (Implementing API version 1.13.2-R0.1-SNAPSHOT)
Dynmap: Dynmap v3.0-beta-5-218

The current workaround for me was to create a custom perspective and set each map a custom perspective.

stormboomer:
linking to https://github.com/leMaik/ChunkyMap/issues/7

FedUpWith-Tech:
I can't replicate this on the latest version of dynmap with Chunky

stormboomer:
@FedUpWith-Tech thank you for having a look at this issue, sadly this problem still exists.

retestet with:
CraftBukkit version 2991-Spigot-018b9a0-f3f3094 (MC: 1.16.5) (Implementing API version 1.16.5-R0.1-SNAPSHOT)
dynmap v3.2-beta-2-483 (from spigot page) and dynmap v3.1-beta7-449 (from release page)
ChunkyMap v2.5.2
Same configuration for worlds as posted above. 

Starting fullrender of world -> render uses chunky renderer to render first map -> online visible tiles = chunky tiles
after the chunky map renderer completes (took about 30 minutes for blank fresh generated world on my test system) the normal dynmap render runs for the seccond map and completes within 20 seconds after the seccond map render is completed the online visible version does not show the chunky tiles anymore but default dynmap tiles. if i look at the chunky tiles while the seccond map render is running the chunky tiles get updated while the seccond map render is still running.

[...]

PR #3475, opened 2021-08-25
Fix for #2564 and Backwards Compatibility for #3436

stormboomer:
Provides a potential fix for #2564 by checking for the provided "renderer" for the specific map.
The fix is to check within `HDMapManager.getShaderStateForTile()` for the "render class" that is used for the map and ignore it if its not comming from org.dynmap --> there probably is a nicer way to do this so I am open to suggestions on a different solution.

While analysing #2564 I noticed that with the latest compiled version from source I was unable to get ChunkyMap running. This is due to the changes made with #3436 which was missing backwards compatiblity for renderes not using the latest tile write Methods.

Furthermore I noticed that by reordering the map configuration of #2564 to
```YML
worlds:
  - name: world
    maps:
      - class: org.dynmap.hdmap.HDMap
        name: surface
        title: Surface Day
        icon: images/block_surface.png
        prefix: surface
        perspective: iso_SE_30_hires
        shader: customshader
        lighting: shadows
      - class: de.lemaik.chunkymap.dynmap.ChunkyMap
        name: chunky
        title: World
        prefix: chunky
        perspective: iso_SE_30_hires
        shader: customshader
        samplesPerPixel: 10
        chunkyThreads: 8
        chunkPadding: 2
        texturepack: texturepacks/texturepack.zip

```
Dynmap stopped the rendering after the "surface" map finished. The ChunkyMap render engine was not started for this map configuration. 
I was able to Fix this by verifiying the "render class" in the Methods `HDMap.getMapsSharingRender()` and `HDMap.getMapNamesSharingRender()` stopping the renderer from falsely interpreting the tiles as rendered in case a different "render class" was used

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/hdmap/HDMap.java
+++ b/DynmapCore/src/main/java/org/dynmap/hdmap/HDMap.java
@@ -46,7 +46,11 @@ public class HDMap extends MapType {
 
     public static final String IMGFORMAT_PNG = "png";
     public static final String IMGFORMAT_JPG = "jpg";
-    
+
+    private String renderClass;
+    public String getRenderClass(){
+        return renderClass;
+    }
     
     public HDMap(DynmapCore core, ConfigurationNode configuration) {
         this.core = core;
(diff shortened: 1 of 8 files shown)


leMaik:
Hi, ChunkyMap maintainer here :wave: 
This issue was fixed in ChunkyMap 2.6.0-pre3, so please give that one a try.

mikeprimm:
I appreciate what Chunky has done, but these interfaces are not part of the API - I'd rather not presume that they are, as I cannot promise that I will not make breaking changes to them in the future.   If the Chunky dev is ready to follow the implementation changes, that would be the best strategy at this point.

Issue #3397, opened 2021-06-20
Depreciated com.mysql.jdbc.Driver


PlayingWithPi:
**Issue Description:**
```yaml
20.06 20:27:52 [Server] INFO [dynmap] Opening MySQL database mysql.apexhosting.gdn:3306/apexMC741534 as map store
20.06 20:27:52 [Server] WARN Loading class `com.mysql.jdbc.Driver'. This is deprecated. The new driver class is `com.mysql.cj.jdbc.Driver'. The driver is automatically registered via the SPI and manual loading of the driver class is generally unnecessary.
```
Using old depreciated driver for MySQL integration, needs to be updated.

* **Dynmap Version:** *v3.1-457**
* **Server Version:** *paper 1.16.5 build 778*
* **Pastebin of Configuration.txt:** *https://pastebin.com/duTGWLRW*
* **Server Host (if applicable):** *Apex Minecraft Hosting*
* **Pastebin of crashlogs or other relevant logs:** [check code snippet above]
* **Steps to Replicate:** *Use MySQL for a storage.*

 [ x ] *I have looked at all other issues and this is not a duplicate*  
 [ x ] *I have been able to replicate this*

mikeprimm-triva:
Loading of the driver via that class is deprecated, but supported.  The driver is not.  We can make the change, but understand that there is zero functional consquence.

patrix87:
Thanks for the clarification, I was here for that too.

generrosity:
@PlayingWithPi @patrix87 - ok to close?

PR #3477, opened 2021-08-28
Add requested Feature to check for MySQL Connector

stormboomer:
Currently Dynmap checks if the class 'com.mysql.jdbc.Driver' exists. For Latest Paper Version this class is marked as deprecated, so this Check will throw an additional error message in the server log.

Luckyly we can just check for  "com.mysql.cj.jdbc.Driver" which is the latest MySQL connector (and supplied by paper).

I have moved the Class check into a function to make it easier to add additional class checks for MySQL Connectors if necessary.

Also related to #3397


mikeprimm (approved):

mikeprimm:
Looks promising - let's give it a go

PR #3480, opened 2021-08-30
fix: explicitly set sign lines

terinjokes:
Minecraft allows for plugins to retrieve the text of a newly created or
modified sign through the `SignChangeEvent.getLines()` method. In
vanilla, the lines returned can be mutated, however a
`SignChangeEvent.setLine(idx, text)` method exists.

PaperMC has internally switched most user-facing text to using the
Adventure Text components, to better support text styling and optimize
serializations. For backwards compatibility with most plugins, PaperMC
retains deprecated `getLines` and `setLine` methods by providing a
legacy serialization view of the Aventure components, however mutations
of the returned list are ignored.

This changeset calls `setLine` after all event listeners ran, allowing
PaperMC to preserve the sign mutations.


mikeprimm:
Thanks for the explanation and the change - let's give this a spin!

PR #3478, opened 2021-08-29
Change radius value of round area markers to int

FedUpWith-Tech:
Should fix #3367

PR #3490, opened 2021-09-18
fix(radius) - gmfamily - process radius as int everywhere (#3478)

gmfamily:
This should fix - globally - the issue #3478. The previous proposed PR is incomplete (only parse radius provided by command as Integer but dont fix the radius representation in all functions -> the world.txt still contains ".0" values at end of all radius).


FedUpWith-Tech:
nice catch!

FedUpWith-Tech (approved):

PR #3500, opened 2021-10-01
Fix GitHub Actions error

inductor:
Hi,

Your GHA has failed due to some changes here the YAML indent was incorrect
https://github.com/rojopolis/spellcheck-github-actions/tree/master#extra-configuration

I also added ForgeGradle in the word list since it was failing in the check.

You can see this is succeeding in https://github.com/inductor/dynmap/pull/1

### DIFF
--- a/.github/workflows/gibberish.txt
+++ b/.github/workflows/gibberish.txt
@@ -52,3 +52,4 @@ gg
 pqBpw
 JDBC
 JDK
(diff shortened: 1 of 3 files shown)


mikeprimm (approved):

mikeprimm:
Looks good - thanks!

PR #3501, opened 2021-10-01
Always run LoadPlayerImages task in the thread pool

JLyne:
Currently `LoadPlayerImages` tasks are only run in the thread pool if `fetchskins` is enabled, otherwise they are immediately run synchronously.

This can become a problem in some situations, such as when Dynmap is configured to store player images in a MySQL database which is currently inaccessible due to temporary network issues. In those cases, the sync task will block the main thread for potentially lengthy periods of time when attempting to save the resulting player images.

This PR removes the `fetchskins` check and always runs the task in the thread pool, which seems harmless.


mikeprimm (approved):

mikeprimm:
Yep - good point - change accepted - thanks!

PR #3467, opened 2021-08-22
Update leaflet version (0.5.1 -> 1.7.1)

JLyne:
This PR aims to update the included leaflet version to the latest (1.7.1). The current version is 8 years old and is missing some notable features and performance improvements. Some of Dynmap's leaflet customisations now have equivalent native features.

Some potential performance improvements this update makes possible with require additional work and can be handled in other PRs, to make this one simpler.

## Notable changes (so far):
- Replaced Dynmap tile size calculation with leaflet's `maxNativeZoom` option
- Removed active layer check as no longer seems necessary
- Replaced custom CRS with default `simple` CRS
- `CustomMarker` now extends `L.Marker`
- Added `CustomIcon` extending `DivIcon`, which handles marker icons
- Updated tiles have a timestamp appended instead of `?upd=0`, allowing tiles to update more than once without caching issues. There are better ways of doing this which can be considered in future PRs.

## Things to do
- [x] Map display
- [x] Layer control
- [x] Regular markers
- [x] Vector layers
- [x] Player markers
- [x] Tile updates

Should fix #3462

Still in progress so things will be broken.

PR #3508, opened 2021-10-04
Fix DynmapTileLayer sharing object instances

### LINKED ISSUE #3506: Visual Bug in Dynmap
* **Dynmap Version:** *3.2-526*
* **Server Version:** *git-Paper-303 (MC: 1.17.1)*
* **Pastebin of Configuration.txt:** *https://pastebin.com/R2ef8Cfx*

Video with bug for easy show what i mean.
https://www.youtube.com/watch?v=LnstuYfQrn8

JLyne:
The leaflet update #3467 introduced a bug where all instances of `DynmapTileLayer` share the same instances of internal objects such as the tile URL cache. This can result in maps with the same name across worlds (such as "flat") using images from the wrong world if the URLs for that world were previously cached.

This PR fixes this by creating internal objects inside an `initialize` method instead of defining them directly in the `.extend` call.

Should fix #3506 and [this comment](https://github.com/webbukkit/dynmap/pull/3504#issuecomment-933160031)

### DIFF
--- a/DynmapCore/src/main/resources/extracted/web/js/dynmaputils.js
+++ b/DynmapCore/src/main/resources/extracted/web/js/dynmaputils.js
@@ -75,10 +75,17 @@ var DynmapLayerControl = L.Control.Layers.extend({
 
 
 var DynmapTileLayer = L.TileLayer.extend({
-	_namedTiles: {},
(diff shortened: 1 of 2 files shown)


mikeprimm:
Thanks...

PR #3516, opened 2021-10-15
Initial pass at 1.18 snapshot support

ajmadsen:
Not currently ready for production. This was my first pass at 1.18 support and I focused on getting surface rendering to look as it did in previous versions (It seems to!). No 1.18 block types have been added yet.

Big changes:
- chunk level NBT changes, tile and biome data moves from previous versions
- (for fabric) 3D biome data now stored at the Section level instead of the Chunk level
- biomes are now encoded by Registry ID and not integer (partially addressed)

I didn't address the Biome change as currently all other versions key biome by an integer and refactoring everything to either use IDs or be backwards compatible was not something I intended on doing before receiving feedback.

As for building and testing, I ran into some issues with `commons-codec` not playing nicely between the mod, fabric, and the snapshot. Currently, for testing outside of gradle, I use the following config in `DynmapCore/build.gradle`:

```
dependencies {
    ...
    implementation 'commons-codec:commons-codec:1.15'
}

shadowJar {
    dependencies {
        ...
        include(dependency('commons-codec:commons-codec:'))
    }
    ...
}
```

### DIFF
--- a/fabric-1.18/src/main/java/org/dynmap/fabric_1_18/ChunkSnapshot.java
+++ b/fabric-1.18/src/main/java/org/dynmap/fabric_1_18/ChunkSnapshot.java
@@ -6,7 +6,9 @@ import net.minecraft.util.collection.PackedIntegerArray;
 import net.minecraft.util.math.MathHelper;
 import net.minecraft.util.math.WordPackedArray;
 import org.dynmap.Log;
+import org.dynmap.common.BiomeMap;
 import org.dynmap.renderer.DynmapBlockState;
+import net.fabricmc.fabric.api.util.NbtType;
 
 import java.util.Arrays;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


ajmadsen:
Actually... let me make this actually diff-able

ajmadsen:
Okay -- the meat of the changes should now be more easily visible instead of a sea of files.

ajmadsen:
I don't know how differing the world storage formats are between different mods, but it seems like it would potentially be useful at this point to pull chunk snapshot code out into a common module, especially as it seems like there are several places that have to check for different storage formats now.

mikeprimm:
Hey - I merged a PR for the GA level (1.18) (https://github.com/webbukkit/dynmap/pull/3539), but I do get the impression that it would benefit all of us to see some of this work integrated with that.
Could you touch base with @kosma or otherwise take a look?   I'm already going to pull in your extensions to BiomeMap, as I plan to use them for the Spigot support!

ajmadsen:
I have to revisit this anyway when I have some time since it looks like some of my assumptions broke in the last snapshot. If it was just block level data changes, @kosma's probably already captures that change. If it's biome related... well I didn't really know how this project wanted to tackle 3D biomes.

kosma:
My PR is pretty similar to yours... there are two unresolved questions,
1. If/how to handle old/new chunk formats
2. How to handle 3D biomes
Especially the second one is big because it touches DynmapCore which is Mike's territory. I don't want to do any 3D biome work in the Fabric ports until Core supports them. Your code is quite useful tho, as it shows how to unpack the biome data - we can hopefully use that!

ajmadsen:
Okay, updated based on the code that already made it to the main branch.

mikeprimm (approved):

mikeprimm:
OK - merged in the new updates -thanks for ironing this out, guys!

Also, my 1.18 priorities right now are:
1) Get Spigot/Paper working
2) Get Forge working
3) Handle any missing block definitions for 1.18 vs the 1.17.1 C&C content I already did
4) Work on core refactor to handle 3D biome data (while also still working with the pre-3D biomes of older Forge and such).
5) Work on shifting chunk NBT processing and biome/block state definition processing to be as common as possible, with chunk data fetching minimized within the platform handlers.

PR #3527, opened 2021-11-12
update hires templates to show actual hires on all maps

### LINKED ISSUE #2834: hires renders 4x4
It is not rendering world 16x16
hires renders 4x4 
4x4
hires

JurgenKuyper:


mikeprimm:
As we discussed in Discord, this will cause impacts on existing folks, since these are the default flat maps for everyone with the default settings.  Whatever is done, it'll need to be something that doesn't invalidate 95% of the flat maps of existing users.

PR #3539, opened 2021-11-30
Fabric 1.18 support

kosma:
Base Fabric 1.18 support:

Changes:
* Remapped to yarn-1.18 mappings
* Changed chunk parsing code to use new chunk format

TODO:
* Support for loading both old and new chunk formats
* Support for 3D biomes (at the moment biome data is ignored)


mikeprimm:
Thanks!  I'll give it a look

PR #3558, opened 2021-12-11
dependency: Update SkinsRestorer

stepech:
SkinsRestorer once again updated their API, which makes Dynmap crash on launch if user enables SkinsRestorer integration in config. Thanks @generrosity for bringing my attention to this.

By SkinsRestorer bStats, less than 3.5% of servers run currently supported SkinsRestorer version (It's even hidden in "others"), all others have newer versions with 92+% of servers running version that crashes dynmap. SkinsRestorer enabled new API in version 14.1, and (probably) disabled old in 14.1.6.

I've just bumped the version and kept already implemented checking for when SkinsRestorer is enabled, but not present, or when version older than 14 is installed. It will work nicely with 14.1+, but unfortunately, there is no easy way to check for 14.0 version (Dynmap will display default skins there, and some SkinsRestorer's stack trace in console is present), but again - that's just a few instances thanks to their aggresive auto updater. 

This PR was tested on Paper 1.18 with several SkinsRestorer versions, including pre v14, some v14.0 and several v14.1 (pre and post 14.1.6)

### DIFF
--- a/spigot/build.gradle
+++ b/spigot/build.gradle
@@ -5,7 +5,7 @@ repositories {
         url 'https://jitpack.io'
     }
     maven {
-        url 'https://repo.codemc.org/repository/maven-snapshots'
+        url 'https://repo.codemc.org/repository/maven-releases/'
     }
 }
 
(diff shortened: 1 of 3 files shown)


stepech:
Will add another catch so that Dynmap won't crash next time SkinsRestorer devs decide to change things around.

mikeprimm (approved):

mikeprimm:
Looks good - let's see how it goes

PR #3561, opened 2021-12-12
Fix boosted top-down-view map glitches

CylonicRaider:
Presently, when using the resolution boosting option on a top-down HDMap (such as one using a `iso_S_90_*` perspective), various blocks/fluids — including, in particular, water — are rendered in a glitchy manner.

This is due to the pixel-centering offset of `0.5` being added *after* performing boost-related coordinate scaling. Because of that, certain raytracing rays end up at integer-valued coordinates, which the intersection testing code does not react well to. (Also, high-resolution tiles may be irritatingly misaligned w.r.t. block boundaries.)

By adding `0.5` in pixel space instead, these issues are resolved.

I estimate this change to be very low-risk (unless the behavior change is deemed undesirable).

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/hdmap/IsoHDPerspective.java
+++ b/DynmapCore/src/main/java/org/dynmap/hdmap/IsoHDPerspective.java
@@ -1248,8 +1248,8 @@ public class IsoHDPerspective implements HDPerspective {
         for(int x = 0; x < tileWidth * sizescale; x++) {
             ps.px = x;
             for(int y = 0; y < tileHeight * sizescale; y++) {
-                ps.top.x = ps.bottom.x = xbase + ((double)x)/sizescale + 0.5;    /* Start at center of pixel at Y=height+0.5, bottom at Y=-0.5 */
-                ps.top.y = ps.bottom.y = ybase + ((double)y)/sizescale + 0.5;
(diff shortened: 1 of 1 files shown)


mikeprimm:
PRETTY sure this is OK - we'll give it a try! :)

FedUpWith-Tech:
@mikeprimm unfortunately this didn't fix the issue-
Here's a cactus outside a boost:true area:
image

Here's a cactus inside a boost:true area:
image

Both are set to png quality. The boosted area has no boostzoom applied yet but simply the boost:true flag set in the marker area. Its harder to tell but the quality of the sand around the cactus is also lower.

I'm waiting on it to render now to see if this behavior changes once the boostzoom is set

FedUpWith-Tech:
Interestingly enough once boostzoom is set the artifacting goes away:
image
image

mikeprimm:
Setting boostzoom doesn't do anything until you rerender....   notice the pixel grain between the two....  boostzoom with an old tile will get stretched by the web browser using whatever method the browser defaults to for stretching under-resolution images.

FedUpWith-Tech:
That was with boost:true set and re-rendered but not boostzoom set. 
After boostzoom was set and re-rendered the artifacting went away.

I guess that's not really an issue since if you are using boost: true you should also be using boostzoom but was not expected behavior for sure

PR #3595, opened 2022-01-04
Cleanup Fabric Gradle projects

kosma:
Generic cleanup, updating the Gradle projects into something more modern and less cluttered. Lots of reduntant code and comments removed. After this cleanup all Fabric `build.gradle` files are the same, paving way for further unification. Additionally, I've updated all Fabric tooling to the newest available versions, fixing *tons* of random issues related to setting up a development environment.


mikeprimm (approved):

mikeprimm:
Let's give this a go - looks promising, but really need to do some test driving to be sure!

PR #3599, opened 2022-01-06
update tiles storage to use mediumblob

JurgenKuyper:
set type to mediumblob to allow for boost type 3 to be stored


mikeprimm (approved):

mikeprimm:
Looks good - trying to decide if we should do an ALTER to handle existing DBs

molor:
So existing DBs will be updated automatically?

molor:
Excellent support :/

_note_: `sarcasm`

JurgenKuyper:
> Looks good - trying to decide if we should do an ALTER to handle existing DBs

That might be good as it shouldn't make a difference for the user

PR #3601, opened 2022-01-07
Fabric 1.18: temporarily revert unification-breaking changes

kosma:
Background of this PR:

1. I am working on merging all the Fabric ports into a single codebase.
2. However, I have also introduced some 1.18-only code that's interfering with the unification effort.
3. Therefore, I'd like to revert those changes until unification is complete - and then reapply them to the unified codebase. That way we can get all the new functionality (fixed `ongenerate` trigger, signs support) on all Fabric versions, not only 1.18.


kosma:
Closing in favor of another one.

PR #3647, opened 2022-02-03
Include t.newImage after db structure changes

DartCZ:


mikeprimm (commented):
Oops - good catch!

mikeprimm:
Thanks for this - good catch!

JurgenKuyper:
shouldn't it be both newImage and Image? to retain the old tiles

mikeprimm:
Yep - I already did follow up fix to address that

Issue #3677, opened 2022-02-13
IP subnet in trusted proxy list


Foxite:
**Feature Description:** Allow ip subnets to be specified in the list of trusted proxies in configuration.txt. For example, 192.168.0.0/24 or 172.0.0.0/8.

* **Additional context:** I need this because my minecraft server is running inside a docker container, and the proxy server is likewise running inside a docker container on the same machine. The minecraft server will see the request coming from an IP address such as 172.0.16.1, but the exact IP address is unpredictable as it may change with every restart of the proxy server container. I would like to simply add 172.0.0.0/8 to the list of trusted proxies, however a cursory reading of the [source code](https://github.com/webbukkit/dynmap/blob/v3.0/DynmapCore/src/main/java/org/dynmap/servlet/SendMessageServlet.java#L95) tells me this won't work.

If it does please tell me and close this issue.

Edit: Fixed "ip range" -> "ip subnet"

jondycz:
Hello Foxite, I see you made changes to support IP ranges. While you're at it, would you mind taking a look at this? Unless I missed something, it's a minor code change in the same file: https://github.com/webbukkit/dynmap/issues/3678

generrosity:
@Foxite you are contributing? awesomesause :) 💚 consider popping onto the Discord #dev chat and talk with the folks there. Mike is often quite happy to include contributers code - tho he does tend to look into the far future when it comes to adding any kinds of new dependancies

mikeprimm:
Hey - I've got this implemented at this point (without adding any dependencies, which are generally problems for mods unless they get shaded and rebased, since they can cause conflict).  My change should also deal with #3678

mikeprimm:
OK - code is in for 3.4-SNAPSHOT (build 750 or later) - please give it a look

Documented in https://github.com/webbukkit/dynmap/blob/v3.0/spigot/src/main/resources/configuration.txt#L413

Foxite:
The feature works in the snapshot version. Thanks!

PR #3687, opened 2022-02-19
IP subnet in trusted proxy list

Foxite:
Implements #3677. This lets the user specify IP subnets such as 192.168.1.0/24 and 172.0.0.0/8 in the trusted proxy list. To do this, I have added a dependency on `com.github.seancfoley:ipaddress:5.3.3` which performs the necessary logic.

Wrote this a few days ago, I'm going to test it now on all supported platforms/versions. Which I think is probably unnecessary but it says so in the contributing document.

Until that's done, perform a code review at your leisure.


mikeprimm:
I've actually implemented this, so you probably don't need to invest more time.  Thanks for the suggestion!

PR #3702, opened 2022-03-05
(add) Feature to change root path if needed

codingric:
Ability to run the web server under a sub path URI


JurgenKuyper:
That is already possible by changing webpath and tilespath in configuration?

mikeprimm:
Not seeing purpose of this, and it will cause a number of other problems

PR #3703, opened 2022-03-06
Small style improvements to marker labels

davidjpfeiffer:
I noticed that marker labels in the web client were not vertically centered, as seen in the below screenshot:

Screen Shot 2022-03-05 at 3 17 21 PM

I made some small style improvements so that the labels are now vertically centered with their icons. I also increased the horizontal padding for the labels from `2px` to `6px` which looks just a little nicer, as you can see in the below screenshot:

Screen Shot 2022-03-05 at 3 35 37 PM

The icons in the two screenshots are different but that is completely unrelated to this change.

Let me know if you have any questions or if there is anything else that I need to update with this change 😃


mikeprimm:
Thanks!

PR #3769, opened 2022-06-02
Fabric 1.19

kosma:


fabric-1.19/gradle.properties
```diff
@@ -0,0 +1,4 @@
+minecraft_version=1.19-pre3
+yarn_mappings=1.19-pre3+build.4
+loader_version=0.14.6
+fabric_version=0.53.4+1.19
```
ThalusA:
```suggestion
minecraft_version=1.19
yarn_mappings=1.19+build.1
loader_version=0.14.6
fabric_version=0.55.2+1.19
```

fabric-1.19/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,33 @@
...
+    ]
+  },
+  "mixins": [
+    "dynmap.mixins.json"
+  ],
+
+  "depends": {
+    "fabricloader": ">=0.14.6",
+    "fabric": ">=0.53.4",
+    "minecraft": "1.19.x"
```
ThalusA:
```suggestion
    "fabricloader": ">=0.14.6",
    "fabric": ">=0.55.2",
    "minecraft": "1.19.x"
```

fabric-1.19/src/main/resources/dynmap.mixins.json
```diff
@@ -0,0 +1,19 @@
+{
+  "required": true,
+  "minVersion": "0.8",
+  "package": "org.dynmap.fabric_1_19.mixin",
+  "compatibilityLevel": "JAVA_16",
```
ThalusA:
```suggestion
  "compatibilityLevel": "JAVA_17",
```
You now require JAVA_17 with 1.19 right ?

fabric-1.19/build.gradle
```diff
@@ -0,0 +1,60 @@
+plugins {
+	// TODO: switch to a stable Loom release once 1.19 lands
+	id 'fabric-loom' version '0.12-SNAPSHOT'
```
ThalusA, line 3: 0.12-SNAPSHOT uses the latest version of 0.12 which means all futures patches (don't change the API at all) 
TODO might be not needed

kosma: I really prefer to pin dependency versions to avoid the build breaking :)

PR #3757, opened 2022-05-12
fix: support special symbols in world names

leon0399:
Fix #3755

Issue #3755, opened 2022-05-12
World name with special symbols breaks marker loading


leon0399:
## **Issue Description:** 
If World Name have special symbols in it's name, seems like frontend trims it's name before first special symbol

| Devtools requests | `fabricworlds.yml` |
|:--------------------| :--------------------|
| ![image](https://user-images.githubusercontent.com/1759654/168124023-844928e4-9b39-4382-b765-202901b77bef.png) | ![image](https://user-images.githubusercontent.com/1759654/168124385-f75bcdad-7901-4546-a1bf-98786365f9f9.png) |

## **How to replicate**
1. Create new world in singleplayer and add special symbols into it's name (at least space and hash (`#`) in my case)
2. Transfer this world's save onto server

## **Server Info**
* **Dynmap Version:** `Dynmap-3.4-beta-3-fabric-1.18.2`
* **Server Version:** Fabric 1.18.2
* **Pastebin of Configuration.txt:** *untouched*, Screenshots above
* **Server Host (if applicable):** Selfhosted
* **Pastebin of crashlogs or other relevant logs:** Screenshots above
* **Other Relevant Data/Screenshots:** *I'm using this texture pack and these plugins and my map is from alpha .8*
* **Steps to Replicate:** *Please be clear in this section so we can replicate your issue*

- [x] *I have looked at all other issues and this is not a duplicate*  
- [x] *I have been able to replicate this*

PR #3771, opened 2022-06-06
feat: URI-encode markers loading world name

leon0399:
One more big, that was unnoticed before by me
Related to #3755 and #3757


mikeprimm:
Looks good

PR #3783, opened 2022-06-21
updated deprecated interfaces where possible

JurgenKuyper:
found some notions of deprecated interfaces, of which the one I could I updated.


forge-1.14.4/src/main/java/org/dynmap/forge_1_14_4/DynmapPlugin.java
```diff
@@ -253,7 +241,10 @@ public void initializeBlockStates() {
                 	if (statename.length() > 0) {
                 		statename += ",";
                 	}
-                	statename += p.getName() + "=" + bs.get(p).toString();
+                    try {
+                        statename += p.getName() + "=" + bs.get(p).toString();
+                    } catch (IllegalFormatConversionException e){
```
ayan4m1, line 246: This should be logged, no?

JurgenKuyper: Could be yes, I only looked at depreciation, not the whole code

mikeprimm (approved):

mikeprimm:
Thanks for update

PR #3784, opened 2022-06-21
Mud Brick fix in texture_1.txt

JurgenKuyper:
fixed mud_bricks sides so it is no longer transparent. thanks BlargCraft on Reddit


kdkasad:
It seems like only the last commit should actually be part of the pull request. The other ones don't seem to be related to the problem the PR is trying to fix.

JurgenKuyper:
true, but had to do a rebase on my repo, because I had a stale commit still hanging and had to pull the other commits in. after fandangling with git for a while found it to work finally.

kosma:
You accidentally committed merge markers to your branch..

JurgenKuyper:
> You accidentally committed merge markers to your branch..

yeah  I saw, fixed it now, am pretty new to all the git stuff, what should I do next time to not have this happen?

kosma:
At this point I'd suggest do a fresh checkout of the upstream repo and redo your changes. Or, alternatively, you can `git cherry-pick` commits between branches in case you have some other work commited but don't want to include it in a PR.

kosma:
Also - if you want to learn how to merge - probably install some good merge tool like Meld, and configure git to use it - it's much safer than doing it by hand. And the final protip - it's a good idea to run `git diff --cached` before you commit, to see exactly what will be in the commit.

JurgenKuyper:
okey thanks, I think I now seem to have fixed it by creating a new branch with remote as base, edited my change, and pushed to the commit. Thanks for the help! I will play around with git some more.

kdkasad:
Still doesn't look quite right. You still have the merge markers in the commit. Also there are a few hundred lines changed when the PR is only talking about fixing one thing, which only takes one line.

Git can definitely be hard, especially when trying to undo something like this. :)

JurgenKuyper:
I don't know what is wrong, I essentially recreated what I did before. also, those merge markers are that the +/_ change signs in the texture_1.txt? which seems odd as I only ever edited the 1 mud_bricks line. And to be fair I am at a loss as to how to fix it.

DynmapCore/src/main/resources/texture_1.txt
```diff
@@ -3201,3 +3201,538 @@ block:id=%attached_melon_stem,patch0=0:melon_stem,patch1=0:attached_melon_stem,b
 block:id=%pumpkin_stem,patch0=0:pumpkin_stem,blockcolor=foliagebiome,transparency=TRANSPARENT,stdrot=true
 # Melon stem
 block:id=%melon_stem,patch0=0:melon_stem,blockcolor=foliagebiome,transparency=TRANSPARENT,stdrot=true
+<<<<<<< HEAD
+=======
```
kdkasad: These kinds of lines are merge markers. Basically, you performed a merge at some point and Git couldn't figure out how to automatically merge the changes, so it put these markers there to tell you that you need to do it manually.

I'd suggest just checking out the upstream master branch, changing the one line, committing it, and pushing that commit to your `brick_fix` branch (using the `--force` flag).

Or if you'd like, I can open a different PR with the changes so you don't have to keep messing around with Git.

JurgenKuyper: I'd rather figure out how to do it, seems like something did change now, and it doesn't show conflicts anymore. So I guess it is now fine to merge?

kdkasad (approved):
Looks good now!

mikeprimm (approved):

mikeprimm:
Thanks for the fix!

PR #3794, opened 2022-07-07
Fix and improve async chunk load

masmc05:
I didn't see some mistakes made by me in last pr, and didn't see that paper has an improved method to save chunk data without many lags
Changes:
1. by mistake i put the check to refuse working with loaded chunks, only from disk, fixed that
2. readded the chunk limit (it's not strict, but still will prevent too much chunk saves, didn't add to chunk loads because it already has lowest priority, so the server will do the load only when server and other plugins don't need that, don't see a necesity)
3. used the paper's method to save the data asyncronously (don't remember why i didn't notice that it has it before)


mikeprimm (approved):

mikeprimm:
Looks good - thanks for the update!

PR #3821, opened 2022-08-08
Bump minimum spigot version, improve armor points counting

stepech:
I've noticed that armor points seem to be counted weirdly by dynmap:

dynmap original full netherite
dynmap original, netherite and elytra

I would expect parity with minecraft client, which doesn't seem to be here. This PR bumps minimum spigot version to 1.10.2 (which is also minimum version officialy supported) from 1.7.10. It also uses new API method introduced in 1.9, that gets precise armor points from bukkit api.

New:

dynmap new full netherite
dynmap new netherite and elytra

Now to be clear.. There was some weird math going on in Armor.java, so I might be touching something I don't understand, but this just seem right, as those past results were unexpected.

----
I also updated SkinsRestorer to check for compatibility issues and to keep dynmap up to date.


stepech:
Old method to get armor durability was also deprecated in 1.13.

mikeprimm:
Thanks!

PR #3841, opened 2022-08-25
Use mojang mappings 

masmc05:
Didn't get a responce in discord if i shoudln't do it (even after the ping), so i did it
I checked in final jar that we use exactly same classes, methods and fields as before (because of that we use `DefaultedRegistry` Instead of `Registry`, even if `DefaultedRegistry` doesn't have it's own static fields and by using it it's same as using `Registry`, but on compiled code they have very diferent names, so to have same compiled code, i decided to use `DefaultedRegistry`)

Accidentally reformated code in 2 classes to use 4 spaces instead of tabs (it was half in that way half in other way, first mistake i made all tabs, then remembered you told standart is 4 spaces), but wish this won't be that bad 

If you accept this then in next pr i'll move other versions to mojang mappings

### DIFF
--- a/bukkit-helper-119/src/main/java/org/dynmap/bukkit/helper/v119/BukkitVersionHelperSpigot119.java
+++ b/bukkit-helper-119/src/main/java/org/dynmap/bukkit/helper/v119/BukkitVersionHelperSpigot119.java
@@ -51,7 +51,7 @@ import java.util.Map;
 /**
  * Helper for isolation of bukkit version specific issues
  */
-public class BukkitVersionHelperSpigot119 extends BukkitVersionHelper {
+public class BukkitVersionHelperSpigot119 extends BukkitVersionHelper { 
     private final boolean unsafeAsync;
 
     public BukkitVersionHelperSpigot119() {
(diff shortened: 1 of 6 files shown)


masmc05:
Oh, yeah, forgot, to compile you need to install the paper api and paper mojang api and delete their .module files

masmc05:
execute 
```
git clone https://github.com/PaperMC/Paper.git
cd paper
./gradlew applyPatches
./gradlew build publishToMavenLocal
```
and then these files needs to be deleted (adapt to your maven local location)
`C:\Users\masmc\.m2\repository\io\papermc\paper\paper-mojangapi\1.19.2-R0.1-SNAPSHOT\paper-mojangapi-1.19.2-R0.1-SNAPSHOT.module`
`C:\Users\masmc\.m2\repository\io\papermc\paper\paper-api\1.19.2-R0.1-SNAPSHOT\paper-api-1.19.2-R0.1-SNAPSHOT.module`
after that it should be able to compile
if you could upload that instead of spigot to your repo, that would be wonderful

bukkit-helper-119/src/main/java/org/dynmap/bukkit/helper/v119/BukkitVersionHelperSpigot119.java
```diff
@@ -199,245 +194,242 @@ public MapChunkCache getChunkCache(BukkitWorld dw, List<DynmapChunk> chunks) {
...
+		BlockEntity tileent = (BlockEntity) te;
+		return tileent.saveWithId();
+	}
+
+	@Override
+	public Object getFieldValue(Object nbt, String field) {
+		CompoundTag rec = (CompoundTag) nbt;
+		Tag val = rec.get(field);
+        if(val == null) return null;
+        if(val instanceof ByteTag) {
```
rautamiekka, line 348: The space usage gets inconsistent as of these statements cuz of `if (` -> `if(`, and other unnecessary extra spaces were added before these lines.

masmc05: I do 0 changes in spaces here, if mike will want these lines to be pretty, he can do that himself, maybe you meant other line but in method getFieldValue there are 0 changes in space usage

masmc05: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference
try this and set default tab size to 4 spaces, then it'll be the same if you really want

masmc05:
Additional build steps not allowed :(

Issue #3839, opened 2022-08-23
Unwanted chunks rendered outside defined limits with limit visibility enabled


gmfamily:
> template is **bold**
> sample data is *italicized*

**Issue Description:** *Some unwanted chunks are rendered outside defined limits with limit visibility enabled. The unwanted chunk are located only around two visible limits defined near to x=20000 z=-20000. We have *

* **Dynmap Version:** *3.4*
* **Server Version:** *paper 1.19 build 81*
* **Pastebin of Configuration.txt:** *https://pastebin.com/8drbHTJa*
* **Server Host (if applicable):** *MineStrator*
* **Pastebin of crashlogs or other relevant logs:** *https://pastebin.com/crashcausedbydynmap*
* **Other Relevant Data/Screenshots:** *No specific texture pack / Plugins (29): ArmorStandTools, AsyncWorldEdit, AsyncWorldEditBossBar, Chunky, CoreProtect, dynmap, Essentials, EssentialsAntiBuild, EssentialsDiscord, EssentialsProtect, EssentialsSpawn, HeadDatabase, HolographicDisplays, LuckPerms, MessageAnnouncer, Multiverse-Core, NoMobGriefing, PlaceholderAPI, ProtocolLib, Skript, spark, SurvivalInvisiframes, TerredesHorizons, Vault, VaultChatFormatter, WorldEdit, WorldEditSelectionVisualizer, WorldGuard, WorldGuardExtraFlags* / Current worlds.txt file : https://pastebin.com/4UAU0UiT / Current dmap worldgetlimits world results : https://pastebin.com/AJE0dwA4 / Online dynmap scoped on problematic limits rendered with unwanted chunks : http://137.74.100.108:25575/?worldname=world&mapname=flat&zoom=1&x=19976&y=64&z=-19994
* **Steps to Replicate:** *dynmap cancelrender world ; dynmap purgeworld world ; dynmap fullrender world -> this sequence repeated 2 times produced exactly the same result with unwanted chunks rendered outside defined limits.*

 [X ] *I have looked at all other issues and this is not a duplicate*  
 [X ] *I have been able to replicate this*

JurgenKuyper:
you have a lot of areas defined, maybe some are overlapping, you can check by visualising the areas with circlemarkers: `dmarker addcircle "label" radius:150 world:world x:-801 y:64 z:975` for example. dynmap sets the areas defined last as the most important ones, so whatever is on top is how it is getting rendered.

[...]

PR #3844, opened 2022-08-31
fix(chnk visibility) Use long to avoid int overflow

gmfamily:
Use long representation of the distance between tested chunk and center of tested limit to avoid int overflow while computing the distance compared to limit radius using square delta value
fix:https://github.com/webbukkit/dynmap/issues/3839


mikeprimm:
Good catch - how big a world did you need to encounter this?

JurgenKuyper:
I think it was like bigger than 20k*20k world if memory serves me right

mikeprimm:
Just curious - been on 100k+ worlds for a long while now: I'd kind of think we'd need to get into that range on both dimensions to actually overflow, and of course it's particular to circular visibility limits :)

gmfamily:
Hello, thank you so much for support and feedback 😃 
The world map is online there: terredeshorizons.minesr.com:25575/?worldname=world&mapname=flat&zoom=0&x=0&y=64&z=0

PR #3857, opened 2022-10-01
Stringify json for json file client async

Warriorrrr:
Makes it so that writeUpdates stringifies the json for the update object async using a CompletableFuture.


mikeprimm:
Looks good - thanks

PR #3861, opened 2022-10-10
Update used paper chunk api

masmc05:
As https://github.com/PaperMC/Paper/pull/8177 was merged, we need to use proper new api, and old api may be removed later


mikeprimm:
Looks good - thanks!

PR #3862, opened 2022-10-14
Colorful biomes

masmc05:
Adds support for 
image
1. support for datapack colors for grass and foliage
2. support for biome grass color modifiers like darker dark forest's color or changeable swamp's color (so supress https://github.com/webbukkit/dynmap/pull/3832 since it adds same effects automatically)

Also add colors for missing biomes (like badlands)
Before:
image
After:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,4 +37,3 @@
 /fabric-1.16.1_server.launch
 /fabric-1.16.2_client.launch
 /fabric-1.16.2_server.launch
-/spigot/run/
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@
 /fabric-1.16.1_server.launch
 /fabric-1.16.2_client.launch
 /fabric-1.16.2_server.launch
+/spigot/run/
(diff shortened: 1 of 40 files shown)


masmc05:
[world.zip](https://github.com/webbukkit/dynmap/files/9785818/world.zip)
the world from screenshot if someone wants to make tests and don't want to search for biomes with custom colors (in vanilla format, but bukkit can transform to bukkit format)

masmc05:
Added support for even 1.16.5 and 1.16.4, but don't like the reflection so won't add for 1.16.2 and 1.16.3, most users won't notice that anyway since they'll use latest 1.16.* already

masmc05:
Tested and works on every version on every platform
if you're ok to download a jar from stranger you can get compiled verion [here](https://drive.google.com/drive/folders/1TRWYzfriwW-oDqKAVM0Q3DKo4fvVmjD9?usp=sharing), but i recommend compiling yourself

mikeprimm:
Looks good - thanks!

PR #3863, opened 2022-10-15
Fix colours

masmc05:
Fix colors in biomes, https://github.com/webbukkit/dynmap/pull/3862 does that automatically, but for older versions this would be needed (if you don't like the more green variant of color, to use the second color that swamp is using, the color needs to be changed to #6a1c24)
New colors: 
image
image


mikeprimm:
Thanks!

Issue #3854, opened 2022-09-28
Missing textures for some Oh The Biomes You'll Go blocks


IMBArator:
Hey there,

I found something strange. I'm trying to integrate Dynmap in a modpack for me and some friends.
But it seems Dynmap or Dynmapblockscan handles some of the blocks in "Oh The Biomes You'll Go" wrong.

There are missing Textures for a bunch of blocks ... but not all. I have already tried to fidle with ``renderdata/``-definitions, but had no success.

In a simple test setup I could identify some specific examples:

* byg:mahogany_leaves
* byg:overgrown_stone
* byg:green_mushroom_block
* byg:white_mushroom_stem

I also could find a example leave block from "Oh The Biomes You'll Go" without any rendering problems:

* byg:ebony_leaves

## Versions

For my test setup I installed the folowing Minecraft and mods on my client there is not need to run a stanalone server to reproduce this:

* Minecraft (1.18.2)
* Forge (40.1.73 or 40.1.80)
* Dynmap (3.4-828)
  * https://www.curseforge.com/minecraft/mc-mods/dynmapforge/files/3934691
* Dymapblockscan (3.4-SNAPSHOT-229)
  * http://dynmap.us/builds/DynmapBlockScan/DynmapBlockScan-3.4-SNAPSHOT-forge-1.18.jar
* "Oh The Biomes You'll Go" (1.4.4)
  * https://www.curseforge.com/minecraft/mc-mods/oh-the-biomes-youll-go/files/3956521
* Terrablender (1.2.0.126)
  * https://www.curseforge.com/minecraft/mc-mods/terrablender/files/3957976

## Configurations

All configs are default. Nothing needs to be changed to replicate.

## How to Replicate

1. Install Minecraft with the above versions
2. Start Minecraft and Create a new world
3. Search a "Oh The Biomes You'll Go"-Biome
  * Some trees have no leaves but others do
  * Overgrown blocks are rendered as black void

### Rendered map with missing textures
![flat](https://user-images.githubusercontent.com/25098428/192793443-848861ab-c367-4855-83b6-c0fd01074027.PNG)
![surface](https://user-images.githubusercontent.com/25098428/192793470-05ad4ad0-c283-4db1-a533-2989d675643e.PNG)

### Blocks with Problems
![not_rendered-mahogany_leaves](https://user-images.githubusercontent.com/25098428/192793673-f031be34-1ab7-49fc-ade9-7bfcfab275fc.png)
[...]

PR #3866, opened 2022-10-25
fix locale issue in ModelBlockModelImpl getLine (closes:  #3854)

IMBArator:
add Locale.US to String.format() where float values are used


mikeprimm:
Good catch - thanks!

PR #3867, opened 2022-10-26
Add better favicon and code for progressive web apps, to improve shortcuts and icons.

surfrock66:
Using the icon for dynmap pulled from the patreon/discord, I updated the favicon and related files to produce a more refined Progressive Web App which looks better as a bookmark on mobile operating systems.

I applied this to my own site as a demo, visible here; adding this to the home screen of an android phone or the windows start menu will demonstrate how it works: https://mc.teh3l3m3nts.com:50780/dynmap

It's been built and tested across all the required versions, though my paper server is the only public one for validation.  All others were tested on https://localhost.  Compiled snapshots can be provided if needed; including the forge 1.12.2 build which worked a bit differently than the build instructions laid out, but I still got to run.

If a different icon would be desired I can use that as well, this is a very trivial change to make and would be very simple to modify as needed


mikeprimm:
Looks good - thanks for the thorough testing!

p120ph37:
This PR breaks iOS "Web Clip" functionality (the "fullscreen on apple mobile products" mentioned in the comment in index.html).  Specifically, it removes the `apple-mobile-web-app-capable` item, and it introduces an android-specific `link rel="manifest"` that causes iOS to fail to open the page in fullscreen mode even if the `apple-mobile-web-app-capable` item is re-added.

seema84:
is there a workable way to change the favicon without it being overwritten with every plugin update?

surfrock66:
It would be a bit of a heavy lift to config; it's multiple files linked in a folder, and files in that folder would have to be edited with specific paths.  If you made some sort of config variable that represented a folder name, you'd have to make sure the whole subset of icons was created for each configurable "theme" and checked out; if one was missing or in-addressable you'd get errors.  The other thing is I didn't touch the config file at all, I'm not sure how it works to pull values from the config file into php.

mikeprimm-mirific:
I could probably add a flag to make it so that the icon files were optionally not extracted during the upgrade extraction.  The index.html is necessarily static on this, since it's common across internal vs external servers and such - anything there would wind up being around accessing another published-but-config-specific file to figure out what to do (e.g. get the name of an alternate favicon), which isn't particularly better a solution.

surfrock66:
I was already doing copy/paste of icons each update, as I made some custom renders; I have multiple isometric views so I added a "NE" or "SW" to the icon.  I don't think it's THAT unreasonable if you change icons/images to add that as a step on updates; I would think making a config to point to new folder and filepath would be tough, but a flag to not overwrite changed files on extract could work?  Is it that tough?

PR #3888, opened 2022-12-08
Fabric 1.19.3

kosma:
Does what it says on the tin.


mikeprimm:
Giving it a go - thanks, as always!!

fabric-1.19.3/src/main/java/org/dynmap/fabric_1_19_3/FabricMapChunkCache.java
```diff
@@ -0,0 +1,103 @@
...
+	// Load generic chunk from unloaded chunk
+	protected GenericChunk loadChunk(DynmapChunk chunk) {
+		GenericChunk gc = null;
+        NbtCompound nbt = readChunk(chunk.x, chunk.z);
+		// If read was good
+		if (nbt != null) {
+			gc = parseChunkFromNBT(new NBT.NBTCompound(nbt));
+		}
+		return gc;
+	}
```
masmc05, line 102: Didn't add the new support for custom biome colours, the original version of this file was outdated

kosma: Oh damn, good catch. I'll fix!

PR #3909, opened 2023-01-03
Hotfix: Fixing the version number for skinsrestorer-api

soulwax:
Fixing the version of the skinrestorer-api from relative to fixed.
The latest version of the skinrestorer-api has moved reflection out of the api, into shared. I therefore propose this as a workaround until the breaking change can be worked out properly.


mikeprimm:
Going with https://github.com/webbukkit/dynmap/pull/3910 for this

PR #3910, opened 2023-01-04
Don't use reflections to access SkinsRestorer API

Joo200:
Dynmap doesn't compile anymore because SkinsRestorer moved ReflectionUtil to another package:
https://github.com/SkinsRestorer/SkinsRestorerX/commit/00feb5005577d70fe60af7094b5ca27fcfa1f82b

This change will replace the the reflection with the proper class type.

### DIFF
--- a/spigot/src/main/java/org/dynmap/bukkit/SkinsRestorerSkinUrlProvider.java
+++ b/spigot/src/main/java/org/dynmap/bukkit/SkinsRestorerSkinUrlProvider.java
@@ -5,7 +5,7 @@ import org.json.simple.parser.JSONParser;
 import org.json.simple.parser.ParseException;
 import org.json.simple.JSONObject;
 import net.skinsrestorer.api.SkinsRestorerAPI;
-import net.skinsrestorer.api.reflection.ReflectionUtil;
+import net.skinsrestorer.api.property.IProperty;
 
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


JurgenKuyper:
Is that a different approach to #3909 ?

Joo200:
Yes and no. The approcch from #3909 is to pin the version from SkinsRestorer.
While this will work while compiling that plugin, that pin will not work when SkinsRestorer on the server will be updated (NoClassFound errors incoming). My approach should fix that but pinning the version is a good idea too.

JurgenKuyper:
I'd guess whatever implementation has the least concurrent changes from our end would be the one accepted, I don't think locking the version is a great idea if it can also be done dynamically

soulwax:
Even better approach would be to contact the  Plugin Dev and ask him why he casually introduced such a breaking change.
Edit: I saw you've done it already, but I'm not convinced of his response.

Joo200:
> Even better approach would be to contact the SkinRestorer Plugin Dev and ask him why he casually introduced such a breaking change.

https://github.com/SkinsRestorer/SkinsRestorerX/commit/00feb5005577d70fe60af7094b5ca27fcfa1f82b#commitcomment-95032542

mikeprimm:
Gonna go with this one, but this all goes to one of the reasons I pulled any sort of integrated external mod specific support years ago, in favor of 'Dynmap-XXX' add on plugins: I cannot have Dynmap breaking due to changes outside of it and the core system.   I'll be probably doing the same with these skin handlers - unlike the security handlers, which are intended to provide APIs, they generally don't approach the world as an API provider, and changes like this are as likely to break existing users running the older skin plugin as they are to fix problems with the new one.

PR #3915, opened 2023-01-12
implemented catch to prevent server crash when jar in jar methods are used

JurgenKuyper:
Forge added the option to nest Jar files, which dynmap doesn't support, added a patch so the server no longer crashes when this is the issue, a full implementation to read the data from the nested jars needs to be implemented but I couldn't get it to work.

### DIFF
--- a/forge-1.18.2/src/main/java/org/dynmap/forge_1_18_2/DynmapPlugin.java
+++ b/forge-1.18.2/src/main/java/org/dynmap/forge_1_18_2/DynmapPlugin.java
@@ -1023,12 +1023,25 @@ public class DynmapPlugin
         }
         @Override
         public File getModContainerFile(String name) {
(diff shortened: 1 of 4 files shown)


kosma:
FWIW, I've had the same problem when trying to do JiJ on Fabric. There's gotta be a way to fix this...

JurgenKuyper:
I left my tried final code in comments to fix it, just couldn't get the proper jar:// path out of it, it kept returning the input jarID for some reason, have been trying for a while and decided to just patch and deploy it.

BartDotRun:
Is there any workaround for now?

JurgenKuyper:
Building the jar yourself is the workaround, from my repo

mikeprimm:
I see what was done here, but am unclear as to when the files being fetched would involve jar-in-jar resolution - can I get a specific case where this is so?

PR #3917, opened 2023-01-17
Update markers.js

JurgenKuyper:
fixed issue with standalone webserver marker resolve with spaces in world name, double urlencode changed %20 to %2520 which didn't resolve to the proper address


mikeprimm:
Looks good!

PR #3932, opened 2023-02-08
implemented user configurable start and stop of cave render

### LINKED ISSUE #3929: Option to change the colours of the depth gradient on the cave map
**Feature Description:** A way for users to define the colour gradient used to represent depth on the cave view. 

**Additional context:** I find the default green/blue gradient very subtle and would like a way to make it clearer. [This page](https://github.com/webbukkit/dynmap/wiki/Defining-custom-shaders) from 2021 says "Future Dynmap releases expect to add options for defining the colors used for cave shading.", but I didn't find any more recent mentions of that. If the option has already been added please let me know!

JurgenKuyper:
if not defined, it uses the old behaviour, if one of the new values are changed, it uses the new gradients. closes #3929


mikeprimm:
Good idea - thanks!

PR #3946, opened 2023-03-16
fabric: fix build issues with 1.19.4 port

kosma:
I promise to never break this again :)


mikeprimm:
Looks good - will build it now!

PR #3948, opened 2023-03-20
Cave gradients

JurgenKuyper:
added new shader to shaders.txt


DynmapCore/src/main/java/org/dynmap/hdmap/CaveHDShader.java
```diff
@@ -19,6 +16,8 @@
 import org.dynmap.utils.MapIterator;
 import org.json.simple.JSONObject;
 
+import javax.swing.*;
```
mikeprimm, line 19: This needs to go - not sure if Swing is usable on headless systems.   Also, please avoid import <package>.*, as it makes package resolution less clear, and can create problems when packages are updated (adding new classes that suddenly start conflicting with other existing ones).

mikeprimm:
Sorry - I've already done the updates for the Chunk version, and did an update for the Cave shader before this (and the data version thing isn't OK, since world heights aren't tied to world version - data packs can change ymin and ymax without regard to version).

PR #3950, opened 2023-03-24
Delete hard reference to MCA file format

sofianedjerbi:
Some of us aren't using standard ANVIL file format. (See [Linear](https://github.com/KaiijuMC/Kaiiju/blob/ver/1.19.3/patches/server/0004-Kaiiju-RegionFormat-Linear.patch) or [Slime](https://github.com/InfernalSuite/AdvancedSlimePaper/blob/main/patches/server/0002-Slime-World-Manager.patch))  
It is simple to allow files extensions such as `.linear` and it is safe to delete hard `.mca` reference.  
I also deleted strange whitespaces (?)

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/common/chunk/GenericMapChunkCache.java
+++ b/DynmapCore/src/main/java/org/dynmap/common/chunk/GenericMapChunkCache.java
@@ -54,11 +54,13 @@ public abstract class GenericMapChunkCache extends MapChunkCache {
 		private DynmapBlockState blk;
 		private final int worldheight;
 		private final int ymin;
+		private final int sealevel;
 
 		OurMapIterator(int x0, int y0, int z0) {
 			initialize(x0, y0, z0);
 			worldheight = dw.worldheight;
 			ymin = dw.minY;
+			sealevel = dw.sealevel;
 		}
 
 		@Override
@@ -483,7 +485,16 @@ public abstract class GenericMapChunkCache extends MapChunkCache {
 		public final int getWorldHeight() {
 			return worldheight;
 		}
-
+		@Override
+	    public final int getWorldYMin() {
+			return ymin;
+		}
+	    /**
+	     * Get world sealevel
+	     */
+	    public final int getWorldSeaLevel() {
+	    	return sealevel;
+	    }
 		@Override
 		public final long getBlockKey() {
 			return (((chunkindex * (worldheight - ymin)) + (y - ymin)) << 8) | (bx << 4) | bz;


bukkit-helper/src/main/java/org/dynmap/bukkit/helper/BukkitWorld.java
```diff
@@ -25,7 +25,7 @@ public class BukkitWorld extends DynmapWorld {
     private World.Environment env;
     private boolean skylight;
     private DynmapLocation spawnloc = new DynmapLocation();
-    
+
```
masmc05: not needed

bukkit-helper/src/main/java/org/dynmap/bukkit/helper/BukkitWorld.java
```diff
@@ -82,7 +82,7 @@ public DynmapLocation getSpawnLocation() {
             Location sloc = world.getSpawnLocation();
             spawnloc.x = sloc.getBlockX();
             spawnloc.y = sloc.getBlockY();
-            spawnloc.z = sloc.getBlockZ(); 
+            spawnloc.z = sloc.getBlockZ();
```
masmc05: not needed

bukkit-helper/src/main/java/org/dynmap/bukkit/helper/BukkitWorld.java
```diff
@@ -184,11 +184,11 @@ public MapChunkCache getChunkCache(List<DynmapChunk> chunks) {
             return null;
         }
     }
-    
+
```
masmc05: not needed

bukkit-helper/src/main/java/org/dynmap/bukkit/helper/BukkitWorld.java
```diff
@@ -184,11 +184,11 @@ public MapChunkCache getChunkCache(List<DynmapChunk> chunks) {
...
     public World getWorld() {
         return world;
     }
-    
+
```
masmc05: not needed

bukkit-helper/src/main/java/org/dynmap/bukkit/helper/BukkitWorld.java
```diff
@@ -206,12 +206,9 @@ public int getChunkMap(TileFlags map) {
         if(lst != null) {
             byte[] hdr = new byte[4096];
             for(File rf : lst) {
-                if(!rf.getName().endsWith(".mca")) {
-                    continue;
-                }
                 String[] parts = rf.getName().split("\\.");
                 if((!parts[0].equals("r")) && (parts.length != 4)) continue;
-                
+
```
masmc05: not needed

masmc05 (changes requested):
> All PRs should be as small as they can be to accomplish the feature or fix being supplied

[Read "Contributing to Dynmap's Code"](https://github.com/webbukkit/dynmap#contributing-to-dynmaps-code)

masmc05:
also maybe you could add a check that dynmap used non `.mca` files, and and write 1 time to console that it will try to work with other formats, but they're not supported? i don't think mike will have time to deal with issues with other formats in future

PR #3970, opened 2023-06-08
Fabric 1.20

kosma:
Does what it says on the tin.

One note, though: I have not done anything about the sign sided-ness or editability. Currently every SignEditPacket is forwarded to DynmapCore as-is. I guess it's up to DynmapCore to decide what to do with it. (This is also known as the "hot potato maneuver" :D)


mikeprimm:
Thanks!!!

PR #4005, opened 2023-08-18
Update dynmap_style.css

JurgenKuyper:
patched issue where the chatbar would drop if login or customlink was enabled, but now the custom link and login overlap if they are set to be in the same area, I don't have enough frontend-knowledge to fix this sadly.


mikeprimm:
Looks good - I'll give it a look tonight

PR #4007, opened 2023-08-23
feat: add `readonly` map option

Jameskmonger:
This option will prevent the map from being updated by the renderer. It can be used if you want a "natural" map, i.e. you want to generate a map and preserve it without being updated with users buildings.

- I'm not sure if these changes are all needed, or all in the right place. It works appropriately in my testing, but as discussed on Discord the only way to find related code is by manually searching. Possibly this change would be better suited elsewhere
- I have updated the template `worlds.txt` file to include the changes.
- This also affects /fullrender and /radiusrender, i.e. they have no effect on a `readonly` map.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/MapManager.java
+++ b/DynmapCore/src/main/java/org/dynmap/MapManager.java
@@ -629,10 +629,12 @@ public class MapManager {
                     renderedmaps.addAll(map.getMapsSharingRender(world));
 
                     /* Now, prime the render queue */
-                    for (MapTile mt : map.getTiles(world, (int)loc.x, (int)loc.y, (int)loc.z)) {
-                        if (!found.getFlag(mt.tileOrdinalX(), mt.tileOrdinalY())) {
-                            found.setFlag(mt.tileOrdinalX(), mt.tileOrdinalY(), true);
-                            renderQueue.add(mt);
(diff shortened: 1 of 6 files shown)


mikeprimm:
Not sure I understand the use case: you can already removed all the rendertriggers (preventing automatic render), and just prevent non-admin users from being able to run the /dynmap *render commands...

Jameskmonger:
My use case is:
- I want to prevent re-renders of `world:natural` but keep `world:live` and `world:inhabited` (both protected maps) rendering as usual

Jameskmonger:
I attempted the following:
- Remove write permissions on the tile folder with `chmod` - this led to too much log spam
- Adding `tileupdatedelay`, of varying high values (ranging from 1 week to 60 years), to the map definitions - this didn't seem to have any effect as the maps still updated.

mikeprimm:
Gotcha - the world-specific requirement is good, so let's go forward here :)

Jameskmonger:
ty for the merge :)

mikeprimm:
All good - I've added a couple of fixes and improvements (the new field was not added to the 'save' method for HDMap, so any use of the /dmap commands would lose the setting; and I added readonly to the dmap commands, so no need to manually edit the worlds.txt files to modify the setting).

PR #4012, opened 2023-08-29
Allow addon mods to track when the internal webserver goes online/offline

Thodor12:
## What is added
A new class to the DynmapCoreAPI, which works similar to the DynmapCommonAPIListener, responsible for tracking listener classes who want to know about state changes in the internal web server.

Features 3 callback methods, when the server is started, stopped, or if the internal webserver was completely disabled through the config.

## Use case
In my addon mod I want to be able to modify the web files to inject (mostly css) to re-style the website itself and markers, rather than relying on people having to manually install new things into the website.

This in itself is already possible, however. The only way possible to track the uptime of the web server is to realisitcally check if the web server files are being created on the filesystem. If the internal webserver is disabled this is never going to happen, leaving my mod into essentially a dead cycle which wouldn't ever complete.
If I can prematurely know if the web server is disabled entirely, I don't have to bother to poll if the index.html exists or not.

Secondly the started/stopped events can help in saying when the server (and thus all it's content) is available, at which point I can safely overwrite any content in the files that I want to add.

## Testing
I verified the code works in the latest Forge, Fabric and Spigot builds to be sure. Though realistically nothing can go wrong unless the callback implementations would throw an exception.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapCore.java
@@ -679,6 +679,7 @@ public class DynmapCore implements DynmapCommonAPI {
                 Log.warning("Using external server, but " + JsonFileClientUpdateComponent.class.toString() + " is DISABLED!");
                 webserverCompConfigWarn = true;
             }
+            DynmapWebserverStateListener.stateWebServerDisabled();
         }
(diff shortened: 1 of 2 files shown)


mikeprimm:
Looks promising - I'll give this a review over the weekend

PR #4013, opened 2023-08-30
Improves Render speed of Zoom tiles by using "free" capacity of the Render Threads.

stormboomer:
Works by checking the amount of active Threads processing tiles. If the amount of Active Threads is less than parallelrendercnt we can use the "spare" capacity for zoom processing. We are using a seperate ThreadPoolExecutor to achive this. Also adds one more cancellation check to process cancellations for large zoom out renders (when many tiles needs to be processed). Additionally it will report Zoom Rendering progress.

Just some metrics for better perspective.
This change improves Total render time (Start of fullrender until zoom out rendering is completed) by 22-23%
The only drawback is that under high load situations the avg tile render speed can drop.
In Terms of usability Results the Zoom Renderer will complete faster thus providing potentially better user experience. Overall with the improvements from #4011 we are still a bit faster in overall render speed compared to before.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
+++ b/DynmapCore/src/main/java/org/dynmap/DynmapWorld.java
@@ -21,6 +21,7 @@ import org.dynmap.utils.Polygon;
 
 import java.awt.image.BufferedImage;
 import java.io.IOException;
+import java.util.concurrent.*;
 
 public abstract class DynmapWorld {
     public List<MapType> maps = new ArrayList<MapType>();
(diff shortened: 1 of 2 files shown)


mikeprimm:
Looks promising - will review this weekend

patrix87:
Could this resolve this issue: https://github.com/webbukkit/dynmap/issues/4109

stormboomer:
@patrix87 it would not solve the issue. However since it will cause Zoom tiles to be rendert quicker, the issue might be mitigated.

PR #4015, opened 2023-09-03
implemented hide if spectator

JurgenKuyper:
implemented hide if spectator for all versions, tested and works


mikeprimm:
I like the idea - a bit concerned about adding it unconditionally, since hidden spectators could be used in PVP as 'spotters', for example, and suddenly having them not show on the map would enable that sort of 'exploit'

JurgenKuyper:
Should have thought of that, implementing a user variable will be my next task

JurgenKuyper:
implemented the requested config value, tested on all versions and is working for me

mikeprimm:
Looks good!

PR #4017, opened 2023-09-14
Proper error catching (regarding Quilt)

MuriloGhignatti:
When the mod is loaded by quilt and you try to get the path using this method an IllegalArgumentException is thrown, I think that quilt is not loading the mod as a file so this is why you are getting this error.

Related to #4016


mikeprimm:
@kosma can you give this a look?

PR #4026, opened 2023-10-15
Improve version helper, use data versions instead of strings

Joo200:
Bukkit exposes the magic number `DataVersion` in `Bukkit.getUnsafe().getDataVersion()`. This version is unique to every release and easy to compare.

To minimize errors by parsing the version string I changed the helper method to get the `BukkitVersionHelper` to use this verison. For older versions before 1.13.2 this method is not exposed yet.

The data versions can be found here: https://minecraft.wiki/w/Data_version#List_of_data_versions


mikeprimm:
Not going to accept this - data versions are NOT equal to releases, they are world schema not API, and these guarantee a break every minor release (which often is not the case), making things MORE brittle, not less.

Issue #3973, opened 2023-06-11
Support other S3 compatible APIs


p47-6:
**Feature Description:** Please Support additional to Amazon S3 also compatible services like [Cloudflare R2](https://www.cloudflare.com/de-de/products/r2/) or [MinIO](https://min.io/)

Essentially it should be possible to freely configure the URL.

NoaHimesaka1873:
Seriously, why is this, a very simple but very important issue, isn't fixed for years? S3-compatible storages have been there for *years*.

MODNMETL:
Yes PLEASE! It is hideously expensive for us to even think about using AWS for our dynmap storage. The ability to use other S3 providers would be awesome!

Issue #3691, opened 2022-02-24
Custom S3 URLs


untuned:
**Feature Description:** Allow setting a custom S3 URL in case someone is using a self-hosted solution, an example of this would be [MinIO](https://github.com/minio/minio).

* **Additional context:** Easier in case I don't want to pay the associated fees, or want to update faster than normal without having to worry about API call costs.

blacksoulgem95:
This is standard in pretty much any software that handles S3 connections, you are allowed to set up custom URLs so that you can connect your own cloud provider's S3 compatible software (eg the ones in GCP, Azure, Wasabi, Linode, Vultr...)

TSStudio:
[dynmap](https://github.com/webbukkit/dynmap)/[DynmapCore](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore)/[src](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src)/[main](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src/main)/[java](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src/main/java)/[org](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src/main/java/org)/[dynmap](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src/main/java/org/dynmap)/[storage](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src/main/java/org/dynmap/storage)/[aws_s3](https://github.com/webbukkit/dynmap/tree/v3.0/DynmapCore/src/main/java/org/dynmap/storage/aws_s3)/AWSS3MapStorage.java is the problem.  
The s3 lite dependency they use does not support other endpoints.  
Thus, to support other S3-compatitable services, they need to find an alternative library or rewrite by themselves.  
If you do need to save files through it now, mount the bucket to your local directory.

Noorquacker:
Update on this: there's an s3rework branch that eliminates the s3-lite dependency and supports custom URLs. I hate to ping unnecessarily but @mikeprimm how much of a priority is this for you right now? I don't mean to put pressure but I would also very much enjoy using my MinIO server for Dynmap :)

itaquito:
[...]

PR #4030, opened 2023-11-17
Allow generic S3 endpoints for alternative services

ChimneySwift:
# Overview
This is a possible implementation for a previously mentioned feature (#3691, #3973), which is custom S3-compatible endpoints in order to allow the use of non-AWS block storage. 

Many of these services offer significantly cheaper alternatives to AWS (with backblaze being almost 4x cheaper). 

I understand there was some work on this in the s3rework branch, but I was more interested in having something soon to see if my plan to use block storage is feasible, and this is potentially a good stop-gap until someone has the time to fully rework the s3 implementation.

I haven't performed much testing on this release, but it appears to work with Backblaze B2. I had some issues trying to get it to work with Cloudflare R2, but I'm not sure if this is more of a Cloudflare-related issue.

# Changes
The main blocker for implementing this is the lack of custom endpoint support within the s3-lite version dynmap uses. There is a new version, 0.0.2-SNAPSHOT, which does support custom endpoint URLs and is a mostly drop-in replacement. Though it's not an official release, there has been no movement on that repo for about a year, so that's as good as it gets without a rewrite.

Anyway, in order to implement this I updated the s3-lite package to 0.0.2-SNAPSHOT and I have added an override_endpoint option to manually set the S3 endpoint. If override_endpoint is blank (the default), it is assumed AWS S3 is being used, and the endpoint is derived from the region.

I have released jars for this version on [my fork](https://github.com/ChimneySwift/dynmap/releases) for anyone interested in testing. I have only tested spigot on 1.20.2 thus far. 

I have also cloned the wiki onto my fork, and have updated the AWS S3 page to include documentation for third-party providers.

### DIFF
--- a/DynmapCore/build.gradle
+++ b/DynmapCore/build.gradle
@@ -19,11 +19,11 @@ dependencies {
     implementation 'org.yaml:snakeyaml:1.23'	// DON'T UPDATE - NEWER ONE TRIPS ON WINDOWS ENCODED FILES
     implementation 'com.googlecode.owasp-java-html-sanitizer:owasp-java-html-sanitizer:20180219.1'
     implementation 'org.postgresql:postgresql:42.2.18'
-    implementation 'io.github.linktosriram:s3-lite-core:0.2.0'
-    implementation 'io.github.linktosriram:s3-lite-api:0.2.0'
-    implementation 'io.github.linktosriram:s3-lite-http-client-url-connection:0.2.0'
-    implementation 'io.github.linktosriram:s3-lite-http-client-spi:0.2.0'
-    implementation 'io.github.linktosriram:s3-lite-util:0.2.0'
+    implementation 'io.github.linktosriram.s3lite:core:0.0.2-SNAPSHOT'
+    implementation 'io.github.linktosriram.s3lite:api:0.0.2-SNAPSHOT'
+    implementation 'io.github.linktosriram.s3lite:http-client-url-connection:0.0.2-SNAPSHOT'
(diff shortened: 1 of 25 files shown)


mikeprimm:
This looks promising - I'll give it a proper review this weekend, and merge if it looks good.  I am not set up to confirm this in any meaningful way, and am not planning on spending money to do so, so other than making sure this doesn't break the AWS S3 (which I use routinely for work), this will be otherwise untested.

ChimneySwift:
Sounds good, this was mainly so I could test some options before pursuing further, but it does seem to work. As long as it doesn't break AWS S3, I don't see a reason not to merge.

I'm not sure I will personally end up using it, not every S3-like service is really geared toward hosting map tiles with frequent gets. I've just got [Bunny CDN](https://bunny.net) set up which is working really well (fast and you can replicate in multiple regions for low latency around the world) but unfortunately it's not S3 compatible. At the moment I have a custom python script syncing the tiles up to bunny, but if I want to use this long term I might consider integrating directly into dynmap, even if only for a personal fork.

An interesting note is that the current S3 implementation is more trying to solve storage constraints, however it could also help bandwidth constrained users, and maybe even help improve performance depending on the service. This can be done by pushing tiles to a CDN and then setting the `url.tiles` setting to point to the CDN instead of the server. The website (and `dynmap_world.json`) is then served from the server, while the tiles are served from the CDN.

obsd-guru:
thanks for adding this feature. was looking for it already. if you need some beta testing pls let me know. i am planning to use it with a minio s3 backend.

mikeprimm:
Merging this - will see how it goes!

PR #4036, opened 2023-12-03
Fix SQLite database size keeps increasing

ApliNi:
https://www.sqlite.org/pragma.html#pragma_auto_vacuum

I once had a Dyanmp SQLite database that expanded to 140GB after running for several months, but the valid data in it was less than 40GB. After adding `auto_vacuum = FULL`, re-rendering and running for 2 months, the database size has barely increased.

and no performance impact was noticed.


mikeprimm:
Looks good - thanks!

PR #4100, opened 2024-04-06
Fixed topo offsets

JurgenKuyper:


DynmapCore/src/main/java/org/dynmap/hdmap/TopoHDShader.java
```diff
@@ -26,15 +23,26 @@ public class TopoHDShader implements HDShader {
...
-        /* Load defined colors from parameters */
-        for(int i = 0; i < 256; i++) {
-            fillcolor[i] = configuration.getColor("color" + i,  null);
+
+        if (HDBlockModels.checkVersionRange(core.getDynmapPluginPlatformVersion(), "-1.17.0")){
+            worldheight = 256;
+            fillcolor = new Color[worldheight];   /* Color by Y, must be range of total world height, offset by +64*/
+            /* Load defined colors from parameters */
+            for(int i = 0; i < worldheight; i++) {
+                fillcolor[i] = configuration.getColor("color" + (i - 64),  null); /* need to substract by 64 because Color does not accept <0 indexes*/
```
minimusubi, line 35: this wont work as expected on nether and end worlds, which do not have an expanded height (0-255)

minimusubi (commented):
just a few thoughts:
- looks like your changes for #4064 are here too. however, this wont matter if 4064 is merged first
- this change assumes all 1.17+ worlds are 363 height- the end and nether are not. though i don't think it will error, it might be cleaner to avoid reading the extra color values
- this also wont work for any worlds with larger custom heights (achievable through data packs or mods)
  - if it works the way I assume it does, it may be better to programmatically read the height values instead of using magic numbers (for example, from `MapIterator.getWorldHeight()` and `MapIterator.getWorldYMin()`). this also future-proofs the code for any future world height changes

PR #4102, opened 2024-04-09
implemented user configurable world height, cleaned up commits

JurgenKuyper:
continuation of #4100, now only one commit wrt v3.0

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/hdmap/TopoHDShader.java
+++ b/DynmapCore/src/main/java/org/dynmap/hdmap/TopoHDShader.java
@@ -6,10 +6,7 @@ import java.io.IOException;
 import java.util.BitSet;
 import java.util.List;
 
-import org.dynmap.Color;
-import org.dynmap.ConfigurationNode;
-import org.dynmap.DynmapCore;
-import org.dynmap.MapManager;
+import org.dynmap.*;
 import org.dynmap.common.DynmapCommandSender;
 import org.dynmap.exporter.OBJExport;
 import org.dynmap.renderer.DynmapBlockState;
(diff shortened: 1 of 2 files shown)


DynmapCore/src/main/java/org/dynmap/hdmap/TopoHDShader.java
```diff
@@ -26,15 +23,26 @@ public class TopoHDShader implements HDShader {
...
+        if (HDBlockModels.checkVersionRange(core.getDynmapPluginPlatformVersion(), "-1.17.0")){
+            worldheight = 256;
+            fillcolor = new Color[worldheight];   /* Color by Y, must be range of total world height, offset by +64*/
+            /* Load defined colors from parameters */
+            for(int i = 0; i < worldheight; i++) {
+                fillcolor[i] = configuration.getColor("color" + (i - 64),  null); /* need to substract by 64 because Color does not accept <0 indexes*/
+            }
         }
+        else{
+            worldheight = configuration.getInteger("worldheight", 348);
```
minimusubi, line 38: `384`, not `348` (64 + 320). however, i would caution against this, as it now requires users to manually specify the world height for nether and end worlds

DynmapCore/src/main/java/org/dynmap/hdmap/TopoHDShader.java
```diff
@@ -26,15 +23,26 @@ public class TopoHDShader implements HDShader {
...
+            fillcolor = new Color[worldheight];   /* Color by Y, must be range of total world height, offset by +64*/
+            /* Load defined colors from parameters */
+            for(int i = 0; i < worldheight; i++) {
+                fillcolor[i] = configuration.getColor("color" + (i - 64),  null); /* need to substract by 64 because Color does not accept <0 indexes*/
```
minimusubi, line 42: `i - 64` will be incorrect for end and nether worlds (they still use 0-255)

minimusubi (commented):
in general, im personally not a fan of having to configure the world height and negative adjustment per-world in a config file. i believe it would be a better user experience if we could detect the world height automatically. this is where mikeprimm's input would be valuable

PR #4118, opened 2024-06-13
Port to 1.21 (rc.1) for fabric

petersv5:
This port to fabric 1.21 (rc.1) seems to work for our use cases. It may be useful as a starting point.
It has only been tested on a lightly loaded testing server.


mikeprimm:
Looks like a good start (lots of new blocks to add!) but this is very helpful.  Thanks!

PR #4120, opened 2024-06-15
1.21 support for Bukkit/Spigot

ashduino101:
This pull request adds support for Minecraft 1.21 on Bukkit/Spigot-based servers. It has been tested on a fresh 1.21 Spigot server, as well as one upgraded from 1.20.4 with world and map data. It should be relatively stable, but more testing is advised.


mikeprimm:
Cool - thanks!   Bunch of new blocks to be done, but this is a helpful start!

Issue #4128, opened 2024-06-30
Hosting on S3 - Null control characters at EOF


kiwikodes:
**Issue Description:** *When using S3 to store Dynmap files, the files round up to the nearest 8KB. This results in null characters at the end of every file, and hence, the files cannot run.*

* **Dynmap Version:** 3.7-SNAPSHOT-949
* **Server Version:** *fabric 1.21 (presumably all other versions also)*
* **Other Relevant Data/Screenshots:** 
*Screenshot of S3 file sizes*
![image](https://github.com/webbukkit/dynmap/assets/81617711/6eb2a0e9-c09f-44e7-95da-d414f83fd302)
*Screenshot of null control characters at end of file*
![image](https://github.com/webbukkit/dynmap/assets/81617711/458869b2-0e96-4fab-ab4e-7e07564b9bb2)
**Steps to Replicate:** 
1. Change configuration.txt to host files on an S3 bucket
2. Run Dynmap as per usual, and try to view the static site from S3 bucket

The site runs but complains about all the JS files. The site will just give a completely black screen, no compass or anything. 

I believe this bug was introduced in #4030. For some reason, every instance of `s3.putObject(req, RequestBody.fromBytes(encImage.buf, encImage.len));` had `encImage.len` removed. This causes the whole buffer to be written to S3 (including the unwanted null characters).

mbarneyjr:
I'm running into this issue. Are there any workarounds until this gets fixed?

PR #4129, opened 2024-06-30
Fix for Breaking S3 Changes

kiwikodes:
Undoing some changes made in #4030 that broke support for S3 (see #4128). These changes will now make sure the appropriate section of the buffer is written to the S3 bucket. Hence resolving #4128.


riipeckx:
Any update on this?

riipeckx (approved):

riipeckx:
> Any update on this?

LGTM btw

stewi1014:
Unfortunately this is not the end of issues from #4030 for me at least.

The zoom tile generation thread gets stuck in an infinite loop trying to read from s3, and silently failing without any message. This process generates about 2-3MB/s of constant traffic to s3 while not successfully writing zoom tiles.

I wonder if you're having the same issue?

Putting a breakpoint in processZoomFile reveals that the call to S3Client.getObjectAsBytes is failing.
image

riipeckx:
@stewi1014 My server have been running for ~6hours for now and still rendering and uploading zoom tiles (not using AWS S3, I'm uploading to Cloudflare R2). As I'm not a Java developer at all I can't ensure I will not be facing an issue soon. Keep us updated :)

PR #4143, opened 2024-08-10
Add 1.21.1 support

forkiesassds:
Closes #4141 
Closes #4144

PR #4146, opened 2024-08-13
1.21.1 Support

Axtazer:
Please check what did I do but I think everything is right ^^

I test it and everything work well for me


Axtazer:
There no big change between version, but I'm not bukkit developer so I'm quite lost that why I tried to do something for making a little support of 1.21.1

Axtazer:
Sorry I open and re-close because idk if it's a good idea it's a duplicate of https://github.com/webbukkit/dynmap/pull/4143 but also it look like that I made more things maybe

PR #4154, opened 2024-08-30
Make webregister command a bit nicer

Jatc252:
- Changed some of the text
- Prevent personalised registration info being sent to sender if using webregister argument with themselves


mikeprimm:
Looks good - thanks!

PR #4162, opened 2024-10-28
Fix s3

stewi1014:
The reason for reverting this is that it breaks zoom tile generation for Amazon's s3 in an insidious way.
No errors are reported, and the zoom tile thread infinitely iterates over calling the s3 API,
throwing an error when s3 replies and then swallowing the error and immediately trying again.
This generates about 2-3MB/s of useless traffic to s3 and the immense amount of API calls made incurs a significant cost.

This solution is non-ideal, and doesn't actually address the core issue; using old unmaintained libraries creates bugs, and poor error management turns bugs into insidious problems.

I'd like to see this library removed and the s3 interface refactored, but have stopped using s3 so don't have a reason/the ability to develop further.


stewi1014:
So sorry, I targeted the wrong repo.

Issue #4185, opened 2025-01-22
Dynmap is doing sync chunkloads from region files for map rendering on the main thread.


sennby:
![Image](https://github.com/user-attachments/assets/e572410c-64f6-4e44-ab2f-a97b237f6d6e)

![Image](https://github.com/user-attachments/assets/078c1eac-a61f-4d94-ad2b-56064d732c4a)

Frequently my server is dropping TPS, and almost always when it is serious it is because of dynmap and this same exact thing. Is there anything i can do?

Im on 1.21.1

Can provide a spark to who needs if it is needed.
every day dynmap #1 lagger on my server sparks. including today again.
its doing sync chunk loads from region files, why does this plugin read chunks for map rendering on the main thread

jacob1:
Before 1.21, there was some code that did async chunk loading on Paper. But it hooked deeply into internal APIs in a way that it shouldn't have. For 1.21, Paper refactored their async loading to make it better, and dynmap broke, it falls back to sync chunk loading instead. Same way that it works on Spigot.

I noticed this when I did my PR for 1.21.3 and 1.21.4. It was way beyond my capabilities to re-add it. Someone did leave a comment about what's needed:
https://github.com/webbukkit/dynmap/pull/4164#issuecomment-2498302212
Based on that comment, it shouldn't be too difficult to re-add it. I never tried it.

sennby:
> Before 1.21, there was some code that did async chunk loading on Paper. But it hooked deeply into internal APIs in a way that it shouldn't have. For 1.21, Paper refactored their async loading to make it better, and dynmap broke, it falls back to sync chunk loading instead. Same way that it works on Spigot.
> 
> I noticed this when I did my PR for 1.21.3 and 1.21.4. It was way beyond my capabilities to re-add it. Someone did leave a comment about what's needed: [#4164 (comment)](https://github.com/webbukkit/dynmap/pull/4164#issuecomment-2498302212) Based on that comment, it shouldn't be too difficult to re-add it. I never tried it.

What can i do as a user not a developer to improve this. Or are you saying that its somethign i should build and do myself and use a custom jar i make, if that is the case I think i will have to be looking at moving off dynmap.

jacob1:
I'm just commenting so other developers know what the issue is and how to fix it. For now you already did everything you need to do - open an issue to bring attention to developers.

sennby:
[...]

PR #4164, opened 2024-11-10
Support Spigot/Paper 1.21.3 and 1.21.4

jacob1:
This PR adds support for 1.21.3 (bundles of bravery update) and 1.21.4 (The Garden Awakens). It's my first time doing a plugin update like this, but I tested it on my server and it seems to be working. For 1.21.4, it just gets it running, it doesn't support rendering the new pale garden / wood blocks.

The biggest difference in this compared to base 1.21 is I removed async chunk loading on paper (#3725). Paper did some massive overhauls of chunk loading on their end (PaperMC/Paper@730882fca9cf , PaperMC/Paper@7bd22b1835af). I think they did this prior to 1.21, but it wasn't caught, because it falls back to sync chunk loading if the classes aren't there. I only noticed because Mojang did their own big refactors to chunk loading in 1.21.2.
It would be worth keeping that ... but unless I had a lot of guidance I think that's way outside what I'm capable of redoing. It was already broken since 1.21 anyway.

I checked the mappings docs to get the right mappings. I found a few mistakes in the older helper classes, mostly inconsequential. I went back and fixed those too.

Tested on Spigot 1.21.4, Spigot 1.21.3, Paper 1.21.3, and Spigot 1.21. Let me know of any changes / additionals / removals you want.

### Jar Download
**Edit2:** This PR is merged, so I deleted my jar download. If the official build still isn't released, use the [dev builds](https://dynmap.us/builds/dynmap/), which support the new blocks now too.

PR #4188, opened 2025-02-02
Async chunk loading for spigot/paper 1.21.3/1.21.4

ZockiRR:
This PR brings back async chunk loading for spigot/paper 1.21.3/1.21.4 requested by #4185.

Basically the impl is described here:
https://github.com/webbukkit/dynmap/pull/4164#issuecomment-2498302212

Tested on paper 1.21.4.

Let me know if anything else is needed.

### DIFF
--- a/bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
+++ b/bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
@@ -6,7 +6,9 @@ import net.minecraft.world.level.biome.BiomeBase;
 import net.minecraft.world.level.biome.BiomeFog;
 import net.minecraft.world.level.chunk.Chunk;
 import net.minecraft.world.level.chunk.storage.SerializableChunkData;
+import org.bukkit.Bukkit;
 import org.bukkit.World;
+import org.bukkit.craftbukkit.v1_21_R2.CraftServer;
 import org.bukkit.craftbukkit.v1_21_R2.CraftWorld;
 import org.dynmap.DynmapChunk;
 import org.dynmap.bukkit.helper.BukkitWorld;
@@ -17,7 +19,10 @@ import org.dynmap.common.chunk.GenericMapChunkCache;
 
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
 
 /**
  * Container for managing chunks - dependent upon using chunk snapshots, since rendering is off server thread
@@ -31,6 +36,22 @@ public class MapChunkCache121_3 extends GenericMapChunkCache {
         super(cc);
     }
 
+    @Override
+    protected Supplier<GenericChunk> getLoadedChunkAsync(DynmapChunk chunk) {
+        CompletableFuture<SerializableChunkData> chunkData = CompletableFuture.supplyAsync(() -> {
+            CraftWorld cw = (CraftWorld) w;
+            Chunk c = cw.getHandle().getChunkIfLoaded(chunk.x, chunk.z);
+            if (c == null || !c.q) { //!c.loaded
+                return null;
+            }
+            return SerializableChunkData.a(cw.getHandle(), c); //SerializableChunkData.copyOf
+        }, ((CraftServer) Bukkit.getServer()).getServer());
+        return () -> {
+            NBTTagCompound nbt = chunkData.join().a(); // SerializableChunkData.write
+            return parseChunkFromNBT(new NBT.NBTCompound(nbt));
+        };
+    }
+
     protected GenericChunk getLoadedChunk(DynmapChunk chunk) {
         CraftWorld cw = (CraftWorld) w;
         if (!cw.isChunkLoaded(chunk.x, chunk.z)) return null;
@@ -39,6 +60,13 @@ public class MapChunkCache121_3 extends GenericMapChunkCache {
         SerializableChunkData chunkData = SerializableChunkData.a(cw.getHandle(), c); //SerializableChunkData.copyOf
         NBTTagCompound nbt = chunkData.a(); // SerializableChunkData.write
         return nbt != null ? parseChunkFromNBT(new NBT.NBTCompound(nbt)) : null;
+    }
+
+    @Override
+    protected Supplier<GenericChunk> loadChunkAsync(DynmapChunk chunk) {
+        CraftWorld cw = (CraftWorld) w;
+        CompletableFuture<Optional<NBTTagCompound>> genericChunk = cw.getHandle().m().a.d(new ChunkCoordIntPair(chunk.x, chunk.z)); // WorldServer.getChunkSource().chunkMap.read(new ChunkCoordIntPair(chunk.x, chunk.z))
+        return () -> genericChunk.join().map(NBT.NBTCompound::new).map(this::parseChunkFromNBT).orElse(null);
     }
 
     protected GenericChunk loadChunk(DynmapChunk chunk) {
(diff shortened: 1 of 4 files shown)


bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
```diff
@@ -31,6 +35,31 @@ public MapChunkCache121_3(GenericChunkCache cc) {
...
+    protected Supplier<GenericChunk> getLoadedChunkAsync(DynmapChunk chunk) {
+        CompletableFuture<GenericChunk> genericChunk = CompletableFuture.supplyAsync(() -> {
+            CraftWorld cw = (CraftWorld) w;
+            Chunk c = cw.getHandle().getChunkIfLoaded(chunk.x, chunk.z);
+            if (c == null || !c.q) { //!c.loaded
+                return null;
+            }
+            SerializableChunkData chunkData = SerializableChunkData.a(cw.getHandle(), c); //SerializableChunkData.copyOf
+            NBTTagCompound nbt = chunkData.a(); // SerializableChunkData.write
+            return parseChunkFromNBT(new NBT.NBTCompound(nbt));
```
masmc05: Kinda missed a bit of the impl description, up to `SerializableChunkData.a(cw.getHandle(), c)` needs to be executed on main thread, the part bellow it should be moved into the supplier so it would be executed async

Basically the cf should return SerializableChunkData and not the parsed chunk

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
```diff
@@ -41,6 +70,24 @@ protected GenericChunk getLoadedChunk(DynmapChunk chunk) {
         return nbt != null ? parseChunkFromNBT(new NBT.NBTCompound(nbt)) : null;
     }
 
+    @Override
+    protected Supplier<GenericChunk> loadChunkAsync(DynmapChunk chunk){
+        CraftWorld cw = (CraftWorld) w;
+        CompletableFuture<GenericChunk> genericChunk = cw.getHandle().m().a.d(new ChunkCoordIntPair(chunk.x, chunk.z)) // WorldServer.getChunkSource().chunkMap.read(new ChunkCoordIntPair(chunk.x, chunk.z))
+                .thenApply(opt -> opt.map(NBT.NBTCompound::new).map(this::parseChunkFromNBT).orElse(null));
```
masmc05: Move this to supplier to make sure it's not gonna be executed on main thread

ZockiRR: This is not how async and cf work. All operations chained with "then" will get executed on the thread the load itself is executed. So basically the return value is not that elegant. In theory the cf itself should be returned, but we are bound to the given interface. So moving this to the supplier will result in getting it executed async when the supplier is requested, but why wait when we can start loading it right now?

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
ZockiRR: But why? I can see no obvious reason to not load in on the parallel thread. No need to stress the main thread here.

jacob1:
I loaded this on my server, and it's working fine. Although, I've never had major performance issues, so I can't really compare a before -> after. mspt seemed around the same, still generally low.

My server is Spigot 1.21.4, which confirms it works there too (Spigot didn't have async chunk loading before).

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
masmc05: It's saving data that is really not liking concurrent access, and accessing chunk in async in Minecraft is always calling for issues

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
masmc05: > All operations chained with "then" will get executed on the thread the load itself is executed.

Yep, and that's what you should avoid, it's not a thread that you can say with confidence that you can stress with additional load. And this is the reason why you should move it to supplier, as you'll for sure be in async thread pool that you can load with stuff like parsing the chunk

ZockiRR: Okay, I think I got you. It was a little missleading at first.
You basically mean to move all further works (after loading itself) to the supplier which means to the Dynmap rendering thread. Right?

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
ZockiRR: Okay, which means that cf needs the main thread as executor. I need to get and copy on the main thread. The rest would be done (once again) on Dynmaps rendering thread, which calls the supplier. Right?

ZockiRR:
The PR has been updated. I hope all matters have been addressed properly.

I have not yet been able to test it myself with paper and hope to do so in the next few days.

@jacob1 would be great if you could test it again with spigot. Thanks in advance.

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
masmc05: Yep

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
masmc05: Yep

bukkit-helper-121-3/src/main/java/org/dynmap/bukkit/helper/v121_3/MapChunkCache121_3.java
```diff
@@ -31,6 +36,22 @@ public MapChunkCache121_3(GenericChunkCache cc) {
...
+        CompletableFuture<SerializableChunkData> chunkData = CompletableFuture.supplyAsync(() -> {
+            CraftWorld cw = (CraftWorld) w;
+            Chunk c = cw.getHandle().getChunkIfLoaded(chunk.x, chunk.z);
+            if (c == null || !c.q) { //!c.loaded
+                return null;
+            }
+            return SerializableChunkData.a(cw.getHandle(), c); //SerializableChunkData.copyOf
+        }, ((CraftServer) Bukkit.getServer()).getServer());
+        return () -> {
+            NBTTagCompound nbt = chunkData.join().a(); // SerializableChunkData.write
```
jacob1: The CompletableFuture might return null, so this throws a NullPointerException if chunkData.join() is null. Assign that to a temporary var and return null if the SerializableChunkData, and that will fix it.

jacob1 (changes requested):
Ok I tested it on my server. It threw an exception, I gave a fix. With that fix, it works great.

hmlendea:
Is there a build artifact of this available anywhere? I'd like to test it on my server on which the latest Dynmap release causes lag.

masmc05:
> Is there a build artifact of this available anywhere? I'd like to test it on my server on which the latest Dynmap release causes lag.

It's a bad idea to download jars from random people, but you can easily compile
Here's the explanation: https://github.com/webbukkit/dynmap?tab=readme-ov-file#how-to-build
Basically first download the code (either use git to clone or use the big green "code" button then "Download code") then run gradle to compile (`gradlew.bat setup build` on windows or `gradlew setup build` for rest operating systems)

ZockiRR:
I fixed the NPE mentioned by @jacob1 and tested on paper. Should be good to go now.

mikeprimm:
Looks promising - thanks!!

PR #4214, opened 2025-06-05
fix (S3): truncate byte buffer when putting object

### LINKED ISSUE #4128: Hosting on S3 - Null control characters at EOF

fosscel:
possible duplicate/necrobump of #4129, resolves #4128

as mentioned in #4129, #4030 introduced a bug that uploads the entire buffer, resulting in null bytes being appended to all files.
this reverts that change, as well as changing to call the overloaded `RequestBody.fromString()`.

before the fix:
image

after the fix:
image

access volume looks normal, tested on a fresh world against an initially empty minio bucket:
image

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/storage/aws_s3/AWSS3MapStorage.java
+++ b/DynmapCore/src/main/java/org/dynmap/storage/aws_s3/AWSS3MapStorage.java
@@ -138,9 +138,9 @@ public class AWSS3MapStorage extends MapStorage {
         			s3.deleteObject(req);
         		}
         		else {
-        			PutObjectRequest req = PutObjectRequest.builder().bucketName(bucketname).key(baseKey).contentType(map.getImageFormat().getEncoding().getContentType())
+        		    PutObjectRequest req = PutObjectRequest.builder().bucketName(bucketname).key(baseKey).contentType(map.getImageFormat().getEncoding().getContentType())
         					.addMetadata("x-dynmap-hash", Long.toHexString(hash)).addMetadata("x-dynmap-ts", Long.toString(timestamp)).build();
-        			s3.putObject(req, RequestBody.fromBytes(encImage.buf));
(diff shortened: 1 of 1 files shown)


MoNoLidThZ:
Thank you for providing this fix to the S3 output issue!

I’ve tested it with my own setup, and it resolves the problem with the extra null characters breaking JSON files.

For anyone interested, here's a [live demo](https://s3.califrogmc.online/buckets-mc/index.html) using this patch.

If you’re unable to build the patched version yourself, I’ve also provided a compiled JAR from [my Jenkins CI instance](https://ci.monolidthz.com/job/Dynmap/).

**Disclaimer:** Always be cautious when downloading JAR files from third parties. I’ve tested this build on my own production server, but please verify and use at your own discretion. The build is provided solely for convenience.

fosscel:
@MoNoLidThZ I was dogfooding this and encountered #4217 , which I resolved in #4218 
currently I'm using [this branch](https://github.com/fosscel/dynmap/tree/s3-double-fix) in my fork that has both of these changes incorporated, if you're interested.

mikeprimm:
Looks good - thanks

PR #4218, opened 2025-06-11
fix (s3): use correct continuation token

### LINKED ISSUE #4217: infinite loop of S3 list requests
**Issue Description:** When processing map tiles, an infinite loop of listing objects can be encountered.

* **Dynmap Version:** v3.0+ (since #3689)
* **Server Version:** all versions
here is a screenshot illustrating the phenomenon, with `GET` requests to minio in red, and other request methods in green:
Image
the switch from green to red coincides with the timing of the `Full render of maps... completed` message being displayed.

this bug occurs because of the use of [`getContinuationToken()`](https://github.com/webbukkit/dynmap/blob/3abbca8582eef65a063f99a9dbf474d49e77087f/DynmapCore/src/main/java/org/dynmap/storage/aws_s3/AWSS3MapStorage.java#L410C77-L410C95) rather than `getNextContinuationToken()` (AWS [documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseSyntax))

* **Steps to Replicate:**

configure dynmap to use s3, and add a custom template suffix:
```
deftemplatesuffix: testing
```
and the `normal-testing.txt` to reproduce this:
```
version: 0.20
templates:
  normal-testing:
    enabled: true
    extrazoomout: 0
    maps:
      - class: org.dynmap.hdmap.HDMap
        name: e90
        title: "e90"
        prefix: e90
        perspective: iso_E_90_vlowres
        shader: stdtexture
        lighting: default
        mapzoomin: 0
      - class: org.dynmap.hdmap.HDMap
        name: s90
        title: "s90"
        prefix: s90
        perspective: iso_S_90_vlowres
        shader: stdtexture
        lighting: default
        mapzoomin: 0
```
after running `/dynmap fullrender`, the `e90` render will complete, and then the `s90` render will not begin.

 [x] *I have looked at all other issues and this is not a duplicate*  
 [x] *I have been able to replicate this*

fosscel:
resolves #4217

as mentioned in #4217 , the use of `getContinuationToken()` rather than `getNextContinuationToken()` causes an infinite loop with S3 requests, and fullrenders won't proceed to the next perspective.

I've tested against the example config in that issue and the problem was resolved.

### DIFF
--- a/DynmapCore/src/main/java/org/dynmap/storage/aws_s3/AWSS3MapStorage.java
+++ b/DynmapCore/src/main/java/org/dynmap/storage/aws_s3/AWSS3MapStorage.java
@@ -407,7 +407,7 @@ public class AWSS3MapStorage extends MapStorage {
 	    		}
 	    		if (result.isTruncated()) {	// If more, build continuiation request
 	    	    	req = ListObjectsV2Request.builder().bucketName(bucketname)
-	    	    			.prefix(basekey).delimiter("").maxKeys(1000).continuationToken(result.getContinuationToken()).encodingType("url").requestPayer("requester").build();
+	    	    			.prefix(basekey).delimiter("").maxKeys(1000).continuationToken(result.getNextContinuationToken()).encodingType("url").requestPayer("requester").build();
 	    		}
 	    		else {	// Else, we're done
 	    			done = true;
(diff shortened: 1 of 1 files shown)


mikeprimm:
Looks good - thanks

PR #4228, opened 2025-07-09
Fabric 1.21.7 & Dried Ghast

reusteur73:
Support for Fabric 1.21.7 and dried ghast.

I tested it on Fabric 1.21.7, and it currently looks like this:
[imgur.com/WbxY2RY.png](https://i.imgur.com/WbxY2RY.png)

However, I wasn't able to correctly render the face's smiley.

The block size appears accurate, but I couldn't get the tentacles to display either.


mikeprimm:
Thanks for this!  I'll take a look at the dried ghast and see if I can sort it out.

PR #4235, opened 2025-08-02
Add Fabric 1.21.8 support

reusteur73:
Extend Fabric support to include version 1.21.8 (grouped with 1.21.6–1.21.7).


mikeprimm:
Thanks - just moved to new house, so zero spare time last couple of months.  I'll try to push this out tonight

PR #4259, opened 2026-05-07
Add support for Bukkit/Paper 26.1.2

surfeur5:
## Add support for Bukkit/Paper 26.1.2

This PR adds compatibility support for Bukkit/Paper `26.1.2` by introducing a dedicated helper module and updating the build environment for newer Java class file support.

### Included changes
- Added new module:
  - `bukkit-helper-261-2`
- Added automatic version detection and helper loading
- Added chunk cache and NBT implementations for 26.1.2
- Registered the new helper in Gradle and Spigot build configuration
- Updated ASM dependencies from `9.5` to `9.8`
  - fixes `Unsupported class file major version 69` during `shadowJar`
- Added Foojay Gradle toolchain resolver support

### Testing
Successfully tested on:
- Paper `26.1.2`
- Java `21`

Build command used:
```bash
./gradlew setup build


jacob1:
The Java portion of this is a line-for-line copy of https://github.com/TwentyOneZ/dynmap/commit/eff762b0ba0e0e53904d4e6cef75d52ef332a88d . Did you copy it from there?

Have you tested this yourself? It was reported on Discord that TwentyOneZ's changes cause trees / plants to render in grayscale. There's a different fork out there that also adds 26.1.2 support and doesn't have the gray issue.

surfeur5:
image

No, I didn’t copy it from TwentyOneZ’s fork.
The changes required for 26.1.2 support are fairly small and straightforward, so similar implementations are expected.

Yes, I tested it myself before opening the PR, and I’m not reproducing the grayscale foliage issue reported on Discord.
From what I can tell, that issue seems related to rendering/color handling changes in that fork, while my PR only focuses on 26.1.2 compatibility.

jacob1:
None of that seems right ... getting an exact 100% copy wouldn't happen even if it is straightforward. But being antagonistic won't help. So I'll just point out what I think the leaf color issue is.

Your PR updates two functions [getFoliageColor / getGrassColor](https://github.com/webbukkit/dynmap/pull/4259/changes#diff-d1b70a54051300ce925614a8754980b6ebf9f4d2f5239e8ebbec9ef9b051c5e7R95-R103). I don't think they match the last known version from [1.21.11](https://github.com/webbukkit/dynmap/blob/v3.0/bukkit-helper-121-11/src/main/java/org/dynmap/bukkit/helper/v121_11/MapChunkCache121_11.java#L94-L104) , which had different method calls. Those different method calls could cause the colors to be wrong in some cases. (maybe they are correct in others).

A more accurate version would mirror the method calls in 1.21.11 [like this fork](https://github.com/hoverwars/dynmap/blob/v3.0/bukkit-helper-26-1-2/src/main/java/org/dynmap/bukkit/helper/v26_1_2/MapChunkCache26_1_2.java#L94-L104).

masmc05:
Yeah I agree that this definitely seems like a copy, even exactly same mistakes. This even touches other modules without testing all the affected versions except for 26.1.2

surfeur5:
I don’t appreciate being accused of copying without a proper technical review of the work.

I already explained that I did not intentionally copy another fork. If there are technical issues in the implementation, I was willing to review and fix them. But I wanted the discussion to stay focused on concrete code review points rather than assumptions about intent.

My intention was only to help add Bukkit/Paper 26.1.2 support. If this contribution is not wanted, that’s fine.

Under these circumstances, I’m withdrawing from this contribution, closing the PR, and deleting the associated branch. I do not want this code to be used or modified as part of this PR.

Thanks.

PR #4263, opened 2026-06-30
Fabric/Forge 26.x support

reusteur73:
This PR builds on top of [taranovegor's PR](https://github.com/webbukkit/dynmap/pull/4262) and adds further support for Fabric 26.1+ and Forge 26.1+

Suggestion: Please merge @taranovegor's PR first, this one includes all his changes plus the additions below.

Added on top of @taranovegor's work:

- Fabric 26.x support *(tested Fabric 26.1 and Fabric 26.2)*
- Forge 26.x support *(tested Forge 26.1 and Forge 26.2)*
- <del>New block textures and model/texture definitions for 26.x blocks</del> *(Now part of #4262)*
- Some bug fixes (Foliage and dirt colors, unnecessary biome mixins and dead Yarn-named mixins)

# Showcase

image

image


AuroraDoesCode:
seems to work fine on 26.2 as well, I just edited the json and recomped, threw it on my server to check. everything seems to work fine for me, and im having no errors in the console.

jacob1:
Having a problem with leaf colors in some biomes, like jungles

image

The same problem was in the 26.1 port too. The bright green jungle leaves are from back when it was running 1.21.11, it's only the new 26.1/26.2 versions that seem to pick the wrong jungle colors.

I tried checking it myself but not sure what the problem is. The only thing I can tell is that the mixin to get the foilage color is probably unnecessary, since it can just directly access it. But still does doesn't fix it.

reusteur73:
Yep, thanks for the feedback! The problem was indeed this:

```java
float tmp = bb.getBaseTemperature(), hum = 0.0f; // set humidity later?
```
(_DynmapPlugin.java:340_)

I implemented a fix and removed the unnecessary mixins.

jacob1:
Thanks for the quick fix. I tested and it's working on my test server now. Will be updating my main server today. Probably should have reported it to you sooner, but it took a while for us to notice after updating to 26.1.

I don't know if I ever would have checked that spot you did for the fix lol.

PR #4266, opened 2026-07-08
Added Folia Support.

KeyHeld:
Added Folia Support.


Tsoccerguy3:
I compilied and tested on 26.2 Paper and Spigot . This may run on Folia , but to be considered a PR under the rules , it can not disable any of the other functionality . Does not run on the other platforms.

Caused by: java.lang.IllegalArgumentException: Error initializing dynmap - bukkit version incompatible!

PR #4271, opened 2026-07-20
Dynmap 3.9

Spartan22411:
### Summary
Adds support for Minecraft 26.2 on Paper/Spigot and Fabric, along with the Gradle 9 migration that's required to build the new Paper module (`paperweight-userdev` needs Gradle 9+).

### Block mapping fixes (26.2)
On 26.2, several blocks rendered as solid black tiles on the map with no error in the server log — Dynmap silently renders any block state it has no `texture_1.txt`/`models_1.txt` mapping for. Diffed `blockstates/*.json` between 1.21.7, 1.21.11, and 26.2 to find every new/changed block, then added mappings tagged `[26.2-]` (so they only apply on 26.2+ and can't affect older versions):

- **Sulfur family** (full block, bricks, chiseled, polished, potent, stairs/slabs/walls) and **Cinnabar family** (same set) — mapped the same way the existing `tuff`/`tuff_bricks` family is
- **Sulfur Spike** — mapped as a billboard cross patch, same technique already used for `pointed_dripstone`
- **Golden Dandelion** / **Potted Golden Dandelion** — mapped like the existing `dandelion`
- **Copper Golem Statue** (8 oxidation/wax variants) — no real block model in vanilla (drawn via `BlockEntityRenderer`), approximated as a solid cube using the matching copper-oxidation texture. Flagged as a follow-up: a faithful render would need a custom `CustomRenderer`.
- **Bug fix:** `creaking_heart` was mapped only for its old `active:true/false` property. Mojang renamed/expanded this to `creaking_heart_state:awake/dormant/uprooted` before 26.2, so every `creaking_heart` rendered black. Added the 9 missing `[26.2-]` mapping lines; the old `[1.21.4-]` entries were left untouched for older servers.

Full details in `CHANGES_26_2_BLOCK_MAPPINGS.md`.

### New modules
- `bukkit-helper-26-2/` — Paper/Spigot 26.2 helper, written directly against real (unobfuscated) `net.minecraft.*` classes, since Mojang ships the server unobfuscated as of 26.1+ (no more Spigot mappings reobf layer)
- `fabric-26.2/` — full Fabric mod port for 26.2, using `net.fabricmc.fabric-loom` (Yarn mappings were discontinued as of 26.1, so no remapping step)
- `spigot/Helper.java` — runtime version dispatch now recognizes 26.2 (`(MC: 26.2)` in the version string, with a reflective `Server#getMinecraftVersion()` fallback)

### Build system
- Main build (Spigot, all `bukkit-helper-*`, all `fabric-*`, DynmapCore/API) moved to **Gradle 9.5.1**
- ForgeGradle (used by `forge-1.14.4`–`forge-1.21.11`) hard-rejects Gradle 9, so those 10 modules now live in their own **Gradle 8.14** build under `forge-build/` (mirrors the existing `forge-1.12.2`/`oldgradle` split) — no source duplication, `forge-build/settings.gradle` just points at the existing `forge-1.x` directories
- Shadow plugin swapped from `io.github.goooler.shadow` (Gradle-8-only) to `com.gradleup.shadow` in the main build
- **Bug fix found during migration:** `com.gradleup.shadow`'s `dependency('group::')` wildcard notation silently matches nothing instead of "everything in that group" (no build error). This had dropped Jetty/javax.servlet/jakarta.xml.bind from `DynmapCore`'s jar and bstats from `spigot`'s — both only caught by loading the plugin on a real server (`NoClassDefFoundError`). Fixed by enumerating each artifact explicitly.

### JDK requirements
- Main build: JDK 21 (unchanged)
- `bukkit-helper-26-2`: JDK 25 (handled by the module's own Gradle toolchain)
- `fabric-26.2`: needs the Gradle **daemon itself** on JDK 25 (Fabric Loom limitation) — build it in a separate invocation from the rest of the main build if you need both

### Testing
Verified on a real 26.2 server: new blocks (sulfur, cinnabar, sulfur spike, golden dandelion, creaking heart, copper golem) now render on the map instead of as black tiles.

Please this is make 100% with claude, no intervention from me, i updated my mc server to 26.2 and made dynmap work on this version, i have no idea if what claude did is good or bad, all i know is that it works fine on my server. Survival paperMC 26.2 10gb ram 
  I hope you can use someting from here. there also a plungin folder for testing if needed. 
   if there is nothing good in this just reply and i will delete this and i will not post again.
   Have a great day and, sorry if i wasted your time.


jacob1:
We do have existing PRs - #4263 and #4262 that add 26.2 support and add all the new textures. I checked and the texture support in those is more thorough. Haven't checked the code portion to compare though.

There's a spigot/paper jar compiled by Tsoccerguy3 in the 2nd PR which should work and contain all the new blocks (although I compiled it myself for my server)

Spartan22411:
> We do have existing PRs - #4263 and #4262 that add 26.2 support and add all the new textures. I checked and the texture support in those is more thorough. Haven't checked the code portion to compare though.
> 
> There's a spigot/paper jar compiled by Tsoccerguy3 in the 2nd PR which should work and contain all the new blocks (although I compiled it myself for my server)

There is no need to waste your time on the code part, if you think there is nothing to use from here no need to look further. it works for me so thats all i needed and im waiting for the official release on 26.2 :P . Have a great day and i love this plugin keep it runing <3 love you all

